PR #2, opened 2023-05-04
WIP: Mdl response

JoGSal:


justAnIdentity:
@cobward let's close this? I think we're good only supporting mdl specific presentations following the -7 profile for the foreseeable future.

PR #3, opened 2023-10-09
Refactor in core.

cobward:


Cargo.toml
```diff
@@ -9,29 +9,28 @@ repository = "https://github.com/spruceid/oidc4vp-rs/"
...
-x509-cert = "0.2.3"
-serde_cbor = "0.11.2"
-anyhow = "1.0.71"
-async-trait = "0.1.72"
-reqwest = "0.11.18"
+anyhow = "1.0.75"
+async-trait = "0.1.73"
+base64 = "0.21.4"
+did-web = "0.2.2"
+didkit = "0.6.0"
```
sbihel: I don't think libraries should use didkit, the APIs should instead accept a `DIDResolver` and let the downstream consumer choose what DIDs to support

sbihel:
Could you take this opportunity to rename the crate to `oid4vp` as well?

src/core/authorization_request/parameters.rs
```diff
@@ -0,0 +1,508 @@
...
+                .json()
+                .await
+                .map(ClientMetadata)
+                .context(format!(
+                    "could not parse response from GET '{}' as JSON",
+                    uri.0
+                ));
+        }
+
+        bail!("")
```
sbihel, line 179: https://github.com/spruceid/oid4vp-rs/pull/4

justAnIdentity:
@cobward this looks great. Finally had a look at it 🙈

I'm planning an update and refactor of isomdl-18013-7.
There will probably be some todo's on here that spill out from that update.
Do you want to merge this and have new issues created to take care of those, or would you rather include them in this refactor in one go?

PR #4, opened 2023-11-25
Log metadata error

brianorwhatever:
@cobward I'm not a rust developer so not sure if this is correct at all but this empty error message caused me quite a bit of grief


sbihel:
Just in case it's relevant to you, we're in the middle of a major refactor/update in #3.

sbihel:
Considering the CI is failing for reasons unrelated to this PR, and the fact that we'll soon have #3 merged, I'm going to close this PR. But thank you for raising this issue, we'll fix it in #3.

PR #6, opened 2024-07-31
Update API and support for draft 20

cobward:


todd-spruceid (approved):

justAnIdentity (approved):
Looks great and should be ready for people to start implementing.

I have a few questions though:

- This doesn't yet support OID4VP Error Responses. Do we feel comfortable putting that on the sprucekit implementers? I think it's better if that lives here
- Same for the RequestVerifier trait implementation for client_id_schemes. I think we should have a default trait implementation here, not in the sprucekit wallet. 
- It doesn't look like we're fully ready for PresentationExchange 2.0, with SubmissionRequirementPicks being ignored in the PresentationDefinition parsing.

Should I open issues for these on this repo?

justAnIdentity:
Maybe we should also provide some support structs for creating JARMs

.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[net]
+git-fetch-with-cli = true
```
sbihel: I don't think this is required now that all dependencies are public (or not git at all)

src/core/util/mod.rs
```diff
@@ -0,0 +1,78 @@
...
+use anyhow::Context;
+use anyhow::Result;
+use async_trait::async_trait;
+use http::{Request, Response};
+
+/// Generic HTTP client.
+///
+/// A trait is used here so to facilitate native HTTP/TLS when compiled for mobile applications.
+#[async_trait]
+pub trait HttpClient {
```
sbihel: I think it would be good to rename this `AsyncHttpClient` to avoid any confusion

src/wallet.rs
```diff
@@ -0,0 +1,88 @@
...
+        AuthorizationRequestObject,
+    },
+    metadata::WalletMetadata,
+    response::{AuthorizationResponse, PostRedirection},
+    util::{base_request, HttpClient},
+};
+
+#[async_trait]
+pub trait Wallet: RequestVerifier + Sync {
+    type HttpClient: HttpClient + Send + Sync;
```
sbihel: All these `Send + Sync` will be a problem if we want to update vp_interop

cobward: Discussed, we will leave it for now and address this later if needed.

cobward:
> This doesn't yet support OID4VP Error Responses. Do we feel comfortable putting that on the sprucekit implementers? I think it's better if that lives here

Good point, will add an issue for this.

> Same for the RequestVerifier trait implementation for client_id_schemes. I think we should have a default trait implementation here, not in the sprucekit wallet.

I disagree with this. There's so much custom stuff that goes into handling the client_id_schemes. We do have some "generic" support for `did` and `x509_san_X` schemes, but they aren't there by default since it would add a load of new dependencies on the trait for things like "trusted verifiers".

> It doesn't look like we're fully ready for PresentationExchange 2.0, with SubmissionRequirementPicks being ignored in the PresentationDefinition parsing.

Yep, I'm adding a issue to generally improve our Presentation Exchange support. We need to have some support for parsing and handling the request, either here or in SpruceKit Mobile.

> Maybe we should also provide some support structs for creating JARMs

I will add an issue for this too - but we also need to consider that we might need to use native crypto in order to meet requirements for FIPS 140-X cryptography.

PR #8, opened 2024-08-08
Fix the presentation submission in the oid4vp-rs e2e test.

Ryanmtate:
As part of this fix, upgrading `ssi` dependency to `0.8.1`;

- [x] Upgrade ssi version to 0.8.1; (use ssi sub-crates)
- [x] Construct a `jwt_vp` token;
- [x] Return a `jwt_vp` token instead of the `jtw_vc` token in the Presentation Submission;
- [x]  Validate the Presentation Submission against the Presentation Definition;
- [x] Validate the VP Token flow, see: https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-6.5 and https://identity.foundation/presentation-exchange/spec/v2.0.0/#input-evaluation;

NOTE: Saving the following for another PR:
- Handle JWT encoded authorization request.


src/core/authorization_request/verification/did.rs
```diff
@@ -6,17 +6,29 @@ use crate::core::{
...
-pub async fn verify_with_resolver(
+pub async fn verify_with_resolver<M>(
     wallet_metadata: &WalletMetadata,
     request_object: &AuthorizationRequestObject,
     request_jwt: String,
     trusted_dids: Option<&[String]>,
-    resolver: &dyn DIDResolver,
-) -> Result<()> {
-    let (headers_b64, _, _) = ssi::jws::split_jws(&request_jwt)?;
+    resolver: &VerificationMethodDIDResolver<impl DIDResolver, M>,
```
Ryanmtate: Not sure if this is the correct resolver type to use here.

Possibly change to `impl JWKResolver` type?

src/core/authorization_request/verification/did.rs
```diff
@@ -64,11 +78,12 @@ pub async fn verify_with_resolver(
...
 
-    let jwk = resolve_key(&kid, resolver)
+    let jwk = resolver
+        .fetch_public_jwk(Some(&kid))
         .await
-        .context("unable to resolve verification method from 'kid' header")?;
+        .context("unable to fetch JWK from 'kid' header")?;
 
-    let _: Json = ssi::jwt::decode_verify(&request_jwt, &jwk)
+    let _: Json = ssi::claims::jwt::decode_verify(&request_jwt, &jwk)
```
Ryanmtate: Given the upgrade to `0.8.1`, the `resolve_key` method is removed.

I changed the argument type to `VerificationMethodDIDResolver` since it provides implementations of `DIDResolver` and `JWKResolver`.

That said, this is not working at the moment, resulting in the following error: 

```
thread 'w3c_vc_did_client_direct_post' panicked at tests/e2e.rs:49:58:
called `Result::unwrap()` on an `Err` value: unable to validate authorization request

Caused by:
    0: unable to validate Authorization Request
    1: unable to fetch JWK from 'kid' header
    2: internal error: DID resolution failed: DID method `key` not supported
```

My sense is that `fetch_public_key` isn't the right method to resolve the JWK.

Should I use a DIDResolver `Output` and parse the public key to create an `JWK` instance instead?

src/presentation_exchange.rs
```diff
@@ -94,9 +104,9 @@ pub struct PresentationSubmission {
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
 pub struct DescriptorMap {
     pub id: String,
-    pub format: String, // TODO should be enum of supported formats
+    pub format: VerifiableFormat,
```
Ryanmtate: Test cases work for this replacement.

src/presentation_exchange.rs
```diff
@@ -84,6 +84,16 @@ pub enum ConstraintsLimitDisclosure {
...
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub enum VerifiableFormat {
+    #[serde(rename = "jwt_vc_json")]
+    JwtVcJson,
+    #[serde(rename = "jwt_vp_json")]
+    JwtVpJson,
+    #[serde(rename = "ldp_vc")]
+    LdpVc,
+}
+
```
Ryanmtate: Are there any other formats we'd like to include here?

Should `jwt_vp` and `jwt_vc` be separate options in addition to these?

Ryanmtate: NOTE: I can also stash these changes and apply to another PR if preferable.

src/verifier/client.rs
```diff
@@ -43,17 +43,18 @@ impl DIDClient {
     pub async fn new(
         vm: String,
         signer: Arc<dyn RequestSigner + Send + Sync>,
-        resolver: &dyn DIDResolver,
+        resolver: impl JWKResolver,
```
Ryanmtate: This may be incorrect resolver type. 

I'm assuming the verification method type needs to resolve into a vm-specific key type, not necessarily a JWK.

src/verifier/client.rs
```diff
@@ -43,17 +43,18 @@ impl DIDClient {
...
         ))?;
 
-        let key = ssi::did_resolve::resolve_key(&vm, resolver)
+        let key = resolver
+            .fetch_public_jwk(Some(&vm))
             .await
             .context("unable to resolve key from verification method")?;
 
-        if &key != signer.jwk() {
+        if &*key != signer.jwk() {
```
Ryanmtate: Used the `VerificationMethodDIDResolver` type since checking the JWK is performed, but not sure this is the correct key resolution process, using `fetch_public_jwk` method above.

Throws the following error:

```
thread 'w3c_vc_did_client_direct_post' panicked at tests/jwt_vc.rs:53:10:
called `Result::unwrap()` on an `Err` value: unable to resolve key from verification method

Caused by:
    internal error: DID resolution failed: DID method `key` not supported
```

tests/jwt_vc.rs
```diff
@@ -36,11 +40,14 @@ pub async fn wallet_verifier() -> (JwtVcWallet, Arc<Verifier>) {
...
+
+    let resolver = VerificationMethodDIDResolver::new(DIDJWK);
+
     let client = Arc::new(
-        oid4vp::verifier::client::DIDClient::new(
+        oid4vp::verifier::client::DIDClient::new::<AnyMethod>(
             verifier_did_vm.clone(),
             signer.clone(),
-            DIDKey.to_resolver(),
+            &resolver,
```
Ryanmtate: I changed the `DIDClient` constructor to take `VerificationMethodDIDResolver`. Not sure if this is the correct verification method resolver or setup.

Ryanmtate:
Merge #9 before this.

Cargo.toml
```diff
@@ -25,7 +25,7 @@ serde_cbor = "0.11.2"
 serde_json = "1.0.107"
 serde_qs = "0.12.0"
 serde_urlencoded = "0.7.1"
-ssi = "0.7.0"
+ssi = "0.8.1"
```
cobward: Would be great if you could also refactor to just use the specific sub-crates that we need in this repo to contract the dependency tree, but feel free to do this once everything is implemented and working.

src/core/authorization_request/verification/did.rs
cobward: Yep, you can remove the associated type and use:

```rust
    pub async fn new(
        vm: String,
        signer: Arc<dyn RequestSigner + Send + Sync>,
        resolver: &VerificationMethodDIDResolver<
            impl DIDResolver,
            impl MaybeJwkVerificationMethod
                + VerificationMethodSet
                + TryFrom<GenericVerificationMethod, Error = InvalidVerificationMethod>,
        >,
```

src/core/authorization_request/verification/did.rs
cobward: I think that is just because you are using DID-JWK instead of DID-Key in the test:
image

src/presentation_exchange.rs
cobward: This is fine, but please add `Other(String)` to capture other variants.

src/verifier/client.rs
```diff
@@ -40,20 +47,26 @@ pub struct DIDClient {
 }
 
 impl DIDClient {
-    pub async fn new(
+    pub async fn new<M>(
         vm: String,
         signer: Arc<dyn RequestSigner + Send + Sync>,
-        resolver: &dyn DIDResolver,
-    ) -> Result<Self> {
+        resolver: &VerificationMethodDIDResolver<impl DIDResolver, M>,
```
cobward: Could you remove the associated type using the same trick as above?

src/verifier/client.rs
cobward: As above, I think this is correct, you just need to pass in the correct DID resolver.

cobward (commented):
Great progress so far!

src/json_schema_validation.rs
```diff
@@ -0,0 +1,204 @@
...
+                bail!(
+                    "String length {} is greater than maximum {}",
+                    s.len(),
+                    max_length
+                );
+            }
+        }
+
+        if let Some(pattern) = &self.pattern {
+            // Note: In a real implementation, you'd use a regex library here
```
Ryanmtate: Replace with regex library for pattern matching.

src/presentation_exchange.rs
Ryanmtate: Using the `ClaimFormat` and `ClaimFormatDesgination` types added in #8. Will add the `Other(String)` type to allow for ad hoc types.

src/verifier/client.rs
Ryanmtate: Perfect, that was the issue! Thanks for the tip!

Ryanmtate:
NOTE: This branch includes updates to be merged in #9.

src/presentation_exchange.rs
cobward: Could you also reuse this enum as the key for the maps [here](https://github.com/spruceid/oid4vp-rs/blob/main/src/core/metadata/parameters/wallet.rs#L137)
and [here](https://github.com/spruceid/oid4vp-rs/blob/main/src/core/metadata/parameters/verifier.rs#L8).

Ryanmtate: Commit [1a2af67](https://github.com/spruceid/oid4vp-rs/pull/8/commits/1a2af6719c47bd8827f43504c0048480e8d273db) adds a designation for `jwt_vc_json` and `jwt_vp_json` with the wallet supported types claim format.

Since technically, its a presentation format more so than a claim format, it might make sense to break this into a separate type, or rename `ClaimFormat` to something more generic, possibly `Format` and `FormatDesignation`.

I'm open to suggestions on the naming and whether it makes sense for it to belong where I put it in the above commit.

src/core/object/mod.rs
```diff
@@ -13,7 +13,7 @@ pub struct UntypedObject(pub(crate) Map<String, Json>);
 // TODO: Replace anyhow error type.
 /// A strongly typed parameter that can represent metadata entries or request parameters.
 pub trait TypedParameter:
-    TryFrom<Json, Error = anyhow::Error> + Into<Json> + Clone + std::fmt::Debug
+    TryFrom<Json, Error = anyhow::Error> + TryInto<Json> + Clone + std::fmt::Debug
```
Ryanmtate, line 16: Add TryInto::Error type restriction to use anyhow::Error.

Ryanmtate: Adding the error type will cause conflicts with the default From<T> infallible error type. Leaving off the TryInto::Error type.

src/holder/verifiable_presentation_builder.rs
Ryanmtate, line 1: Considering yanking this. Currently, it is not being used.

src/core/authorization_request/parameters.rs
```diff
@@ -193,7 +193,28 @@ impl TryFrom<Json> for ClientMetadataUri {
...
+        &self.0
+    }
+}
+
+impl Nonce {
+    pub fn random() -> Self {
+        // NOTE: consider replacing with rng rand crate.
+        Self(uuid::Uuid::new_v4().to_string())
+    }
+}
```
Ryanmtate: Replace with a better nonce generator?

src/presentation_exchange.rs
```diff
@@ -1,80 +1,772 @@
...
+    /// Checks the underlying presentation submission parsed from the authorization response,
+    /// against the input descriptors of the presentation definition.
+    pub fn validate_authorization_response(
+        &self,
+        auth_response: &AuthorizationResponse,
+    ) -> Result<()> {
+        match auth_response {
+            AuthorizationResponse::Jwt(jwt) => {
+                bail!("Authorization Response Presentation Definition Validation Not Implemented.")
+            }
```
Ryanmtate: TODO: Handle Jwt encoded authorized response.

src/json_schema_validation.rs
```diff
@@ -0,0 +1,344 @@
...
+/// meets the expected schema.
+///
+/// For more information, see the field constraints filter property:
+///
+/// - [https://identity.foundation/presentation-exchange/spec/v2.0.0/#input-descriptor-object](https://identity.foundation/presentation-exchange/spec/v2.0.0/#input-descriptor-object)
+///
+/// - [https://json-schema.org/understanding-json-schema](https://json-schema.org/understanding-json-schema)
+///
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct SchemaValidator {
```
Ryanmtate: Add a [`const`](https://json-schema.org/understanding-json-schema/reference/const) value for equality check or restricted to a certain value. 

Add [`enum`](https://json-schema.org/understanding-json-schema/reference/enum) type to ensure `one-of` predicate for inclusion in a set.

Ryanmtate: Ensure the validator matches the definitions specified here: https://json-schema.org/draft/2020-12/json-schema-validation

Ryanmtate: This is handled in #9 and has been merged into this branch.

src/presentation_exchange.rs
```diff
@@ -1,80 +1,846 @@
...
+///
+/// The object MUST include a format-specific property (i.e., alg, proof_type) that expresses which
+/// algorithms the Verifier supports for the format. Its value MUST be an array of one or more
+/// format-specific algorithmic identifier references, as noted in the [ClaimFormatDesignation].
+///
+/// See [https://identity.foundation/presentation-exchange/spec/v2.0.0/#presentation-definition](https://identity.foundation/presentation-exchange/spec/v2.0.0/#presentation-definition)
+/// for an example schema.
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(untagged)]
+pub enum ClaimFormat {
```
Ryanmtate: There is a bug when serializing and deserializing the `ClaimFormat` type, resulting in `JwtVcJson` being deserialized into `JwtVpJson`.

This occurs in the e2e test. Re-arranging the enum field order does change the behavior, but does not fix the bug.

Marking this problem as it will have significant side effects and is difficult to detect. So it should be tracked for a bugfix.

src/presentation_exchange.rs
Ryanmtate: Should we resolve this now or in another PR?

src/presentation_exchange.rs
Ryanmtate: This has been resolved. Removing the `untagged` seems to do the trick.

src/presentation_exchange.rs
```diff
@@ -277,6 +319,71 @@ impl PresentationDefinition {
...
+
+                // TODO: Verify the JWT.
+                // let jws = CompactJWSString::from_string(jwt.clone()).context("Invalid JWT.")?;
+                // let resolver: VerificationMethodDIDResolver<DIDKey, AnyJwkMethod> =
+                //     VerificationMethodDIDResolver::new(DIDKey);
+                // let params = VerificationParameters::from_resolver(resolver);
+
+                // if let Err(e) = jws.verify(params).await {
+                //     bail!("JWT Verification Failed: {:?}", e)
+                // }
```
Ryanmtate: TODO: Verify the JWT

src/presentation_exchange.rs
```diff
@@ -386,6 +493,109 @@ impl InputDescriptor {
...
+
+                        // According the specification, found here:
+                        // [https://identity.foundation/presentation-exchange/spec/v2.0.0/#input-evaluation](https://identity.foundation/presentation-exchange/spec/v2.0.0/#input-evaluation)
+                        // > If the result returned no JSONPath match, skip to the next path array element.
+                        continue;
+                    }
+
+                    // If a filter is available with a valid schema, handle the field validation.
+                    if let Some(Ok(schema_validator)) = validator.as_ref() {
+                        // TODO: possible trace a warning if a field is not valid.
```
Ryanmtate: currently using `debug!` to trace filter validation errors. Possibly unnecessary.

src/presentation_exchange.rs
```diff
@@ -386,6 +493,109 @@ impl InputDescriptor {
...
+                                    "Field did not pass filter validation, required by predicate."
+                                );
+                            } else if constraint_field.is_required() {
+                                bail!("Field did not pass filter validation, and is not an optional field.");
+                            }
+                        }
+                    }
+
+                    // TODO: Check limit disclosure of data requested. Do not provide more data
+                    // than is necessary to satisfy the constraints.
```
Ryanmtate: Remove, is handled above starting on line 547.

src/presentation_exchange.rs
```diff
@@ -575,11 +789,47 @@ impl ConstraintsField {
...
 
+    /// Return a JSON schema validator using the internal filter.
+    ///
+    /// If no filter is provided on the constraint field, this
+    /// will return None.
+    ///
+    /// If the filter schema is invalid, this will also return None.
+    ///
+    /// NOTE: Errors are not handled if the filter schema is invalid,
+    /// instead the method will return None on an invalid filter.
```
Ryanmtate: Remove comment. Validation errors are surfaced.

tests/e2e.rs
```diff
@@ -41,47 +57,54 @@ async fn w3c_vc_did_client_direct_post() {
...
+    let descriptor_map = parsed_presentation_definition
+        .parsed()
+        .input_descriptors()
+        .iter()
+        .map(|descriptor| {
+            // NOTE: the input descriptor constraint field path is relative to the path
+            // of the descriptor map matching the input descriptor id.
+            DescriptorMap::new(
+                descriptor.id().clone(),
+                ClaimFormatDesignation::JwtVc,
```
Ryanmtate: Add the following code comment:

```
// NOTE: Since the input descriptor may support several different claim format types. This value should not be 
// hardcoded in production code, but should be selected from available formats in the presentation definition 
// input descriptor.
```

tests/jwt_vp.rs
Ryanmtate, line 1: This file will create and save a new VP Token.

tests/jwt_vp.rs
```diff
@@ -0,0 +1,117 @@
...
+use ssi_claims::vc::v2::syntax::VERIFIABLE_PRESENTATION_TYPE;
+use ssi_claims::{JWSPayload, JWTClaims};
+// use ssi_claims::vc::v1::VerifiableCredential;
+use ssi_claims::jwt;
+use ssi_dids::ssi_json_ld::syntax::{Object, Value};
+use ssi_dids::ssi_json_ld::CREDENTIALS_V1_CONTEXT;
+use ssi_dids::DIDKey;
+use ssi_jwk::JWK;
+
+#[tokio::test]
```
Ryanmtate: Determine if this test should be ignored so it does not overwrite existing vp token.

NOTE: the vp token expires after 1 hour, so any test checking expiration will need to generate a valid token.

tests/jwt_vp.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    let jwt = claim
+        .sign(&holder_jwk)
+        .await
+        .expect("Failed to sign Verifiable Presentation JWT");
+
+    println!("JWT: {:?}", jwt);
+
+    // Save the JWT to the file system to be used in other tests `examples/vp.jwt`
+    std::fs::write("tests/examples/vp.jwt", jwt.as_str())?;
```
cobward: Could we just have this function return it rather than using the FS? I have bad experiences with tests relying on local state :)

src/core/authorization_request/parameters.rs
cobward:
```suggestion
impl Nonce {
    pub fn random(length: u8) -> Self {
        Self(Alphanumeric.sample_string(&mut rand::thread_rng(), length))
    }
}
```

src/presentation_exchange.rs
Ryanmtate: commit [1398cad](https://github.com/spruceid/oid4vp-rs/pull/8/commits/1398cadffd13c2a03fc69fc8514656a7830c3f73) resolves

Ryanmtate:
Once this receives approval, I will squash these commits.

Cargo.toml
```diff
@@ -9,27 +9,28 @@ repository = "https://github.com/spruceid/oidc4vp-rs/"
 documentation = "https://docs.rs/oid4vp/"
 
 [features]
+default = []
 reqwest = ["dep:reqwest"]
 p256 = ["dep:p256"]
+rand = ["dep:rand"]
```
cobward: Unnecessary feature overhead, `rand` is already in the dependency tree through the ssi crates. And `reqwest` is too as of adding `jsonschema`, so we can remove that feature too. `p256` should be optional, but with the ssi refactor it looks like that is always imported through `ssi-dids`, even with --no-default-features. I think that is a bug, but we can remove it as a feature for now.

Cargo.toml
```diff
@@ -46,3 +47,9 @@ uuid = { version = "1.2", features = ["v4", "serde", "js"] }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 uuid = { version = "1.2", features = ["v4", "serde"] }
+
+# For Development Purpose, patch the ssi* crates to use the local versions
+# [patch.crates-io]
+# ssi-claims = { path = "../ssi/crates/claims" }
+# ssi-dids = { path = "../ssi/crates/dids" }
+# ssi-jwk = { path = "../ssi/crates/jwk" }
```
cobward: My preference is to keep this out of the Cargo.toml, as it makes assumptions about the developer's directory structure.

src/core/metadata/parameters/wallet.rs
```diff
@@ -134,7 +137,7 @@ impl From<RequestObjectSigningAlgValuesSupported> for Json {
 
 // TODO: Better types
```
cobward: This TODO comment can be removed I think

src/holder/verifiable_presentation_builder.rs
cobward, line 1: @timothee-haudebourg @sbihel  what do you think about adding something like this to `ssi` instead of here?

src/verifier/mod.rs
```diff
@@ -119,12 +119,12 @@ impl Verifier {
         validator_function: F,
     ) -> Result<()>
     where
-        F: FnOnce(Session, AuthorizationResponse) -> Pin<Box<Fut>>,
-        Fut: Future<Output = Outcome>,
+        F: FnOnce(Session, AuthorizationResponse) -> Fut,
+        Fut: Future<Output = Result<Outcome>>,
```
cobward: This is intentionally not `Result<Outcome>`. If there is an error, it should be captured in the `Outcome::Failure` variant.

src/verifier/client.rs
```diff
@@ -43,17 +44,20 @@ impl DIDClient {
     pub async fn new(
         vm: String,
         signer: Arc<dyn RequestSigner + Send + Sync>,
-        resolver: &dyn DIDResolver,
+        resolver: impl JWKResolver,
     ) -> Result<Self> {
+        // use ssi_jwk::JWKResolver;
```
cobward:
```suggestion
```

tests/e2e.rs
```diff
@@ -8,80 +10,112 @@ use oid4vp::{
...
-    }))
-    .unwrap();
+    let presentation_definition = PresentationDefinition::new(
+        "did-key-id-proof".into(),
+        InputDescriptor::new(
+            "did-key-id".into(),
+            Constraints::new()
+                .add_constraint(
+                    ConstraintsField::new(
+                        "$.vp.verifiableCredential[0].vc.credentialSubject.id".into(),
```
cobward:
```suggestion
                        "$.credentialSubject.id".into(),
```

tests/e2e.rs
cobward, line 1: Could you make something more like (or an exact replica of) this example from the specification?
https://openid.net/specs/openid-4-verifiable-presentations-1_0-20.html#name-w3c-verifiable-credentials

tests/e2e.rs
```diff
@@ -8,80 +10,112 @@ use oid4vp::{
...
+                    .set_predicate(Predicate::Required),
+                )
+                .set_limit_disclosure(ConstraintsLimitDisclosure::Required),
+        )
+        .set_name("DID Key Identity Verification".into())
+        .set_purpose("Check whether your identity key has been verified.".into())
+        .set_format((|| {
+            let mut map = ClaimFormatMap::new();
+            map.insert(
+                ClaimFormatDesignation::JwtVp,
```
cobward:
```suggestion
                ClaimFormatDesignation::JwtVcJson,
```

tests/e2e.rs
```diff
@@ -8,80 +10,112 @@ use oid4vp::{
...
+                // NOTE: Since the input descriptor may support several different claim format types. This value should not be
+                // hardcoded in production code, but should be selected from available formats in the presentation definition
+                // input descriptor.
+                //
+                // In practice, this format will be determined by the VDC collection's credential format.
+                ClaimFormatDesignation::JwtVc,
+                "$".into(),
+            )
+        })
+        .collect();
```
cobward: This doesn't make sense to me, shouldn't we presenting something like:
```json
    {
      "id": "did-key-id",
      "path": "$",
      "format": "jwt_vp_json",
      "path_nested": {
        "path": "$.vp.verifiableCredential[0]",
        "format": "jwt_vc_json"
      }
    }
```

tests/e2e.rs
```diff
@@ -8,80 +10,112 @@ use oid4vp::{
...
+
+    let presentation_submission = PresentationSubmission::new(
+        uuid::Uuid::new_v4(),
+        parsed_presentation_definition.parsed().id().clone(),
```
cobward, line 128: Must the presentation definition id match the presentation submission id? If they _can_ be different then I think it should be done so here in the example.

cobward (changes requested):

Cargo.toml
Ryanmtate: Did we want the ability to pass in a reference to an existing RNG? e.g.

```rust
impl Nonce {
    pub fn random(rng: &mut impl rand::Rng, len: u8) -> Self {
        use rand::distributions::Alphanumeric;

        Self((0..len).map(|_| rng.sample(Alphanumeric) as char).collect())
    }
}
```

I don't know if this is a common case where an RNG is already initialized and could be passed in.

Cargo.toml
Ryanmtate: Do you have another preference for patching libs? Or just remove the commented code from the Cargo.toml and let the developer add it as necessary?

tests/e2e.rs
Ryanmtate: Presentation Submission has its own unique ID, however, the `definition_id` field within the presentation submission must match the originating presentation definition ID in the authorization request.

Cargo.toml
Ryanmtate: I didn't see `p256` as a re-export from `ssi-dids` for `0.2.0`: https://docs.rs/ssi-dids/latest/ssi_dids/#reexports

I also don't see `rand` as a re-export from `jsonschema`: https://docs.rs/jsonschema/latest/jsonschema/#reexports

How are these libraries used from dependencies, if they are not re-exported?

tests/e2e.rs
Ryanmtate: Yeah, do we want to change the `vc.jwt` example payload? I based the Presentation Definition loosely on the example vc, but it is non-normative.

Ryanmtate: Here's the payload of the example vc currently used:

```json
{
  "iss": "did:key:zDnaeeex9MAVbhoWeDcbbGZdzM1zxqZqpC387jWoLhUr1BdST",
  "nbf": 1704067200,
  "jti": "2c282eba-d147-422c-9ce5-991f199800c5",
  "sub": "did:key:zDnaefqT1BrGGsJEZGwAiueouqMh6MqsZhaL1md5hkHgtfzb2",
  "vc": {
    "@context": "https://www.w3.org/2018/credentials/v1",
    "id": "2c282eba-d147-422c-9ce5-991f199800c5",
    "type": "VerifiableCredential",
    "credentialSubject": {
      "id": "did:key:zDnaefqT1BrGGsJEZGwAiueouqMh6MqsZhaL1md5hkHgtfzb2"
    },
    "issuer": "did:key:zDnaeeex9MAVbhoWeDcbbGZdzM1zxqZqpC387jWoLhUr1BdST",
    "issuanceDate": "2024-01-01T00:00:00+00:00"
  }
}
```

Ryanmtate: Commit https://github.com/spruceid/oid4vp-rs/pull/8/commits/9997486535fbf5ffe1c7f4968f280aef4f9f9337 updates the path to be top-level. But I am also wondering myself if this is correct.

It would seem the top level should be the `Verifiable Submission`, which I would think would include a `jwt_vp_json` located at the `vp` verifiable presentation field.

In that case, shouldn't the top-level `jwt_vp_json` be the `$.vp` field? and not at the presentation submission level? Or is the presentation submission assumed to be a `jwt_vp_json` response payload.

I went with your suggestion, because I tried to track down some example cases in the specification, but didn't see anything definitive.

That said, the descriptor map, when constructing the presentation should know the relative path, and it should match one of the constraint field paths from the input descriptor when constructed.

tests/e2e.rs
Ryanmtate: I've updated the example to more closely resemble your suggestion, see: 

https://github.com/spruceid/oid4vp-rs/pull/8/files#diff-62403d73a2340f1d6ac792a3d62161b538ffefa05bc7365f96351854116d63b1R99-R124

NOTE: This is still using the existing `vc.jwt` example credential.

src/presentation_exchange.rs
```diff
@@ -277,6 +320,71 @@ impl PresentationDefinition {
...
+
+                let jws = CompactJWSString::from_string(jwt.clone()).context("Invalid JWT.")?;
+                let resolver: VerificationMethodDIDResolver<DIDKey, AnyJwkMethod> =
+                    VerificationMethodDIDResolver::new(DIDKey);
+
+                let params = VerificationParameters::from_resolver(resolver);
+
+                if let Err(e) = jws.verify(params).await {
+                    bail!("JWT Verification Failed: {:?}", e)
+                }
```
Ryanmtate: NOTE: The e2e test is currently failing due to `MissingPublicKey`.

```
unexpected status: Complete(Error { cause: JWT Verification Failed: MissingPublicKey })
```

I'm assuming I am incorrectly signing the verifiable presentation claim? Is this the correct flow?

https://github.com/spruceid/oid4vp-rs/pull/8/files#diff-6d6c7d16f3d6119451ac6b8c431591a112676ff96645d7a6cba9055aa85a7d2aR95-R100

Ryanmtate: Looking at the companion test wallet, it seems I should be using the `Presentation::generate_jwt` from `ssi_vc` to generate the signed JWT.

This brings up another question: should we bump the version of ssi-vc `0.2.*` -> `0.3.*`?

It looks like `Presentation` has become a trait and not a struct type.

For now, I will use the `0.2.1` version and see if I can get the e2e test working correctly.

src/holder/verifiable_presentation_builder.rs
Ryanmtate, line 1: Should we yank this? It is dead code, and the `holder` module is not used.

I think the helper methods should go directly on the `VerifiablePresentation` `ssi-jwt` claims type.

tests/jwt_vp.rs
```diff
@@ -0,0 +1,51 @@
...
+
+    let verifiable_presentation =
+        VerifiablePresentationBuilder::from_options(VerifiablePresentationBuilderOptions {
+            issuer: holder_did.clone(),
+            subject: holder_did.clone(),
+            audience: verifier_did.clone(),
+            expiration_secs: 3600,
+            credentials: vec![verifiable_credential],
+            nonce: "random_nonce".into(),
+        });
```
Ryanmtate, line 39: Decided to keep the verifiable presentation builder and provide a helper method. We may still want to move this closer to the VerifiablePresentation struct itself, but at least this will provide a helper method for reuse.

src/core/credential_format/mod.rs
```diff
@@ -15,3 +19,248 @@ pub struct JwtVc;
...
+/// algorithms the Verifier supports for the format. Its value MUST be an array of one or more
+/// format-specific algorithmic identifier references, as noted in the [ClaimFormatDesignation].
+///
+/// See [https://identity.foundation/presentation-exchange/spec/v2.0.0/#presentation-definition](https://identity.foundation/presentation-exchange/spec/v2.0.0/#presentation-definition)
+/// for an example schema.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub enum ClaimFormat {
+    #[serde(rename = "jwt")]
+    Jwt {
+        // The algorithm used to sign the JWT.
```
cobward: Please use `///` for all of these comments

src/core/credential_format/mod.rs
```diff
@@ -15,3 +19,248 @@ pub struct JwtVc;
 impl CredentialFormat for JwtVc {
     const ID: &'static str = "jwt_vc";
 }
+
```
cobward: If the old trait isn't used anymore you can delete it.

src/core/input_descriptor.rs
```diff
@@ -0,0 +1,513 @@
...
+        self.format = Some(format);
+        self
+    }
+
+    /// Validate the input descriptor against the verifiable presentation and the descriptor map.
+    pub fn validate_verifiable_presentation(
+        &self,
+        verifiable_presentation: &VerifiablePresentation,
+        descriptor_map: &DescriptorMap,
+    ) -> Result<()> {
```
cobward, line 161: Could you add some positive and negative unit tests for this method?

src/core/presentation_definition.rs
```diff
@@ -0,0 +1,185 @@
...
+    /// against the input descriptors of the presentation definition.
+    pub async fn validate_authorization_response(
+        &self,
+        auth_response: &AuthorizationResponse,
+    ) -> Result<()> {
+        match auth_response {
+            AuthorizationResponse::Jwt(_jwt) => {
+                // TODO: Handle JWT Encoded authorization response.
+
+                bail!("Authorization Response Presentation Definition Validation Not Implemented.")
```
cobward:
```suggestion
                bail!("Authorization Response Presentation Definition validation not implemented.")
```

src/core/presentation_definition.rs
```diff
@@ -0,0 +1,185 @@
...
+                            // TODO: Determine whether input descriptor must have a corresponding descriptor map.
+                            bail!("Input Descriptor ID not found in Descriptor Map.")
+                        }
+                        Some(descriptor) => {
+                            input_descriptor
+                                .validate_verifiable_presentation(
+                                    &verifiable_presentation,
+                                    descriptor,
+                                )
+                                .context("Input Descriptor Validation Failed.")?;
```
cobward:
```suggestion
                                .context("Input Descriptor validation failed.")?;
```

src/core/presentation_definition.rs
```diff
@@ -0,0 +1,185 @@
...
+
+                bail!("Authorization Response Presentation Definition Validation Not Implemented.")
+            }
+            AuthorizationResponse::Unencoded(response) => {
+                let presentation_submission = response.presentation_submission().parsed();
+
+                let jwt = response.vp_token().0.clone();
+
+                let verifiable_presentation: VerifiablePresentation =
+                    ssi_claims::jwt::decode_unverified(&jwt)?;
```
cobward: Rather than assuming that the vp_token is a jwt_vp_json, could you do a match on the expected format first?

src/holder/verifiable_presentation_builder.rs
cobward: Yep let's yank then.

src/lib.rs
```diff
@@ -1,5 +1,7 @@
 pub mod core;
-pub mod presentation_exchange;
+pub mod holder;
```
cobward, line 2:
```suggestion
```

cobward (changes requested):

Cargo.toml
cobward:
```
$ cargo tree -i rand
rand v0.8.5
├── num-bigint-dig v0.8.4
│   └── rsa v0.6.1
│       └── ssi-jwk v0.2.1
...
```

```
$ cargo tree -i reqwest@0.12.5
reqwest v0.12.5
└── jsonschema v0.18.0
```

```
$ cargo tree -i p256
p256 v0.13.2
├── did-method-key v0.2.2
│   [dev-dependencies]
│   └── oid4vp v0.1.0 (/Users/jacob/Documents/git/oidc4vp-rs)
│       [dev-dependencies]
│       └── oid4vp v0.1.0 (/Users/jacob/Documents/git/oidc4vp-rs) (*)
├── did-method-key v0.3.0
│   └── ssi-dids v0.2.0
...
```

cobward: > How are these libraries used from dependencies, if they are not re-exported?

We don't need to use the crates from those dependencies, there's just no point gating them behind features if they're already in the dependency tree.

Cargo.toml
cobward: Yep the latter.

src/verifier/mod.rs
cobward: This still needs to be reverted.

tests/e2e.rs
cobward: We should follow the profile defined in OID4VP:

1. Verifier requests a `jwt_vc_json` with some contraints on that VC.
2. Holder generates a `jwt_vp_json` which is a presentation of that VC, including the transaction's nonce in the VP.
3. Holder presents the VP, identifying where the VC can be found within the VP.

cobward: Just to point out: OID4VP is a framework specification. The semantics are not well defined, and it's down to the "profile" to define how a request is interpreted, and what kind of response to be built.

src/core/metadata/mod.rs
```diff
@@ -89,6 +117,18 @@ impl WalletMetadata {
...
+    /// field from the wallet metadata.
+    pub fn request_object_signing_alg_values_supported(&self) -> Result<Vec<String>, Error> {
+        let Some(Ok(algs)) = self.get::<RequestObjectSigningAlgValuesSupported>() else {
+            bail!(
+                "Failed to parse request object signing algorithms supported from wallet metadata."
+            )
+        };
+
+        Ok(algs.0)
+    }
```
cobward: No need to add this method, this should be done via dereference to UntypedObject.

src/presentation_exchange.rs
cobward: Another PR

src/tests.rs
```diff
@@ -0,0 +1,192 @@
...
+        println!("✅")
+    }
+}
+
+#[test]
+fn test_input_descriptor_validation() -> Result<()> {
+    // Include the `input_descriptors_example.json` file in the `examples` directory.
+    let input_descriptors = include_str!(
+        "../tests/presentation-exchange/test/presentation-definition/input_descriptors_example.json"
+    );
```
Ryanmtate: FYI: Submission Requirement groups is currently not handled in the validation flow.

Working on the `group` submission requirements in the validation flow to make this test work. Using the examples from the test examples git submodule.

src/lib.rs
Ryanmtate: I did end up keeping the `holder` module, but I think it is temporary, and would like to move these builder methods closer to the `VerifiablePresentation` struct in the future.

src/core/presentation_definition.rs
Ryanmtate: I removed the jwt encoding of the vp_token, and followed the docs more closely for a base64 encoded payload, which now handled if multiple VPs are included in the VP Token.


> REQUIRED. JSON String or JSON object that MUST contain a single Verifiable Presentation or an array of JSON Strings and JSON objects each of them containing a Verifiable Presentations. Each Verifiable Presentation MUST be represented as a JSON string (that is a Base64url encoded value) or a JSON object depending on a format as defined in Appendix A of [[OpenID.VCI](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#OpenID.VCI)]. When a single Verifiable Presentation is returned, the array syntax MUST NOT be used. If Appendix A of [[OpenID.VCI](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#OpenID.VCI)] defines a rule for encoding the respective Credential format in the Credential Response, this rules MUST also be followed when encoding Credentials of this format in the vp_token response parameter. Otherwise, this specification does not require any additional encoding when a Credential format is already represented as a JSON object or a JSON string.

see: https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-6.1-2.2

src/core/input_descriptor.rs
Ryanmtate: Working on these tests. I am covering the submission requirement groups, which were previously not handled in the input descriptors. Using the `tests/presentation-exchange` submodule for example presentation definitions and submissions.

src/core/credential_format/mod.rs
Ryanmtate: These are now removed.

src/presentation_exchange.rs
Ryanmtate: I've removed the JWT encoding and the verification from the e2e test. Using the unencoded flow. Will open a new PR for the JWT processing.

tests/e2e.rs
```diff
@@ -1,87 +1,146 @@
...
+            map
+        })()),
+    );
 
     let client_metadata = UntypedObject::default();
 
+    #[cfg(feature = "rand")]
+    let nonce = Nonce::random(&mut rand::thread_rng());
+
+    #[cfg(not(feature = "rand"))]
```
Ryanmtate: TODO: Remove unused cfg feature.

tests/e2e.rs
```diff
@@ -1,87 +1,146 @@
...
+                ClaimFormatDesignation::JwtVcJson,
+                ClaimFormatPayload::Alg(vec![Algorithm::ES256.to_string()]),
+            );
+            map
+        })()),
+    );
 
     let client_metadata = UntypedObject::default();
 
+    #[cfg(feature = "rand")]
```
Ryanmtate: TODO: removed unused cfg feature.

tests/jwt_vc.rs
```diff
@@ -152,10 +159,15 @@ impl AsyncHttpClient for MockHttpClient {
...
-                            info: serde_json::Value::Null,
+                |session, auth_response| {
+                    Box::pin(async move {
+                        match auth_response.validate(&session.presentation_definition) {
+                            Ok(_) => Outcome::Success {
+                                info: serde_json::Value::Null,
+                            },
+                            Err(e) => Outcome::Error {
+                                cause: e.to_string(),
+                            },
```
Ryanmtate: I've moved the `validate` method to the `auth_response` implementation instead of a method of the presentation definition. I think that makes more sense.

src/core/presentation_submission.rs
```diff
@@ -0,0 +1,151 @@
...
+    ///
+    /// The format of a path_nested object mirrors that of a [DescriptorMap] property. The nesting may be any number of levels deep.
+    /// The `id` property MUST be the same for each level of nesting.
+    ///
+    /// > The path property inside each `path_nested` property provides a relative path within a given nested value.
+    ///
+    /// See: [https://identity.foundation/presentation-exchange/spec/v2.0.0/#processing-of-submission-entries](https://identity.foundation/presentation-exchange/spec/v2.0.0/#processing-of-submission-entries)
+    ///
+    /// Errors:
+    /// - The id of the nested path must be the same as the parent id.
```
Ryanmtate: TODO: Remove this error message, we're no longer throwing an error in this method.

src/tests.rs
Ryanmtate: Submission requirement groups are now handled.

src/core/metadata/mod.rs
Ryanmtate: This method has been removed.

src/lib.rs
Ryanmtate: Keeping this for now until the functionality is handled closer to the `jwt::VerifiablePresentation`.

Cargo.toml
```diff
@@ -40,7 +40,7 @@ x509-cert = "0.2.4"
 serde_path_to_error = "0.1.8"
 tokio = { version = "1.32.0", features = ["macros"] }
 did-method-key = "0.2"
-oid4vp = { path = ".", features = ["p256"] }
+oid4vp = { path = "." }
```
cobward:
```suggestion
```

src/core/credential_format/mod.rs
```diff
@@ -1,17 +1,291 @@
...
+        proof_type: Vec<String>,
+    },
+    #[serde(rename = "ac_vp")]
+    AcVp {
+        /// The proof type used to sign the anoncreds verifiable presentation.
+        proof_type: Vec<String>,
+    },
+    #[serde(rename = "mso_mdoc")]
+    MsoMDoc(serde_json::Value),
+    Other(serde_json::Value),
```
cobward, line 84: I think this `Other` case needs `#[serde(untagged)]` otherwise it will only capture:
```json
{
  ...
  "Other": { ... }
  ...
}
```

Could you add a little unit test for deserializing and the `designation` fn for this case?

src/core/credential_format/mod.rs
```diff
@@ -1,17 +1,291 @@
...
+    /// The format is defined by ISO/IEC 18013-5:2021 [ISO.18013-5](https://identity.foundation/claim-format-registry/#term:iso.18013-5)
+    /// which defines a mobile driving license (mDL) Credential in the mobile document (mdoc) format.
+    /// Although ISO/IEC 18013-5:2021 ISO.18013-5 is specific to mobile driving licenses (mDLs),
+    /// the Credential format can be utilized with any type of Credential (or mdoc document types).
+    #[serde(rename = "mso_mdoc")]
+    MsoMDoc,
+    /// Other claim format designations not covered by the above.
+    ///
+    /// The value of this variant is the name of the claim format designation.
+    Other(String),
```
cobward, line 220: As above I think this needs a `#[serde(untagged)]`. Please add a test.

src/core/credential_format/mod.rs
```diff
@@ -1,17 +1,291 @@
...
+    ///
+    /// Given there is no universal standard for how to present a vehicle title credential,
+    /// the inner String provides a dynamic way to represent a vehicle title credential.
+    // TODO: is there a standard identifier for a vehicle title credential instead of `vehicle_title`?
+    #[serde(rename = "vehicle_title")]
+    VehicleTitle(String),
+    // TODO: Add additional credential types here. Fallback to a string for any other credential type.
+    /// Other credential types not covered by the above.
+    Other(String),
+}
```
cobward: We will never be able to enumerate all credential types, so I think a string is the most reasonable representation.

src/core/credential_format/mod.rs
```diff
@@ -1,17 +1,291 @@
...
-    const ID: &'static str;
+use std::collections::HashMap;
+
+use serde::{Deserialize, Serialize};
+
+// TODO: Does the `isomdl` crate provide this constant value, or another crate?
+const ORG_ISO_18013_5_1_MDL: &str = "org.iso.18013.5.1.mDL";
+
+/// A Json object of claim formats.
+pub type ClaimFormatMap = HashMap<ClaimFormatDesignation, ClaimFormatPayload>;
```
cobward, line 6: Why is this needed in addition to `ClaimFormat`? Isn't `ClaimFormatMap` equivalent to `Vec<ClaimFormat>`?

src/core/input_descriptor.rs
```diff
@@ -0,0 +1,610 @@
...
+pub struct InputDescriptor {
+    id: String,
+    #[serde(default)]
+    constraints: Constraints,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    name: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    purpose: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    format: Option<ClaimFormatMap>,
```
cobward: Using [KeyValueMap](https://docs.rs/serde_with/latest/serde_with/struct.KeyValueMap.html), you can use `Vec<ClaimFormat>` instead of `ClaimFormatMap` for better typed values.

src/core/input_descriptor.rs
```diff
@@ -0,0 +1,610 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    group: Option<Vec<GroupId>>,
```
cobward: Is it important to distinguish between `None` and an empty vec/map? If not:
```suggestion
    #[serde(default, skip_serializing_if = "ClaimFormatMap::is_empty")]
    format: ClaimFormatMap,
    #[serde(default, skip_serializing_if = "Vec::is_empty")]
    group: Vec<GroupId>,
```

src/core/input_descriptor.rs
```diff
@@ -0,0 +1,610 @@
...
+
+/// Constraints are objects used to describe the constraints that a [Holder](https://identity.foundation/presentation-exchange/spec/v2.0.0/#term:holder) must satisfy to fulfill an Input Descriptor.
+///
+/// A constraint object MAY be empty, or it may include a `fields` and/or `limit_disclosure` property.
+///
+/// For more information, see: [https://identity.foundation/presentation-exchange/spec/v2.0.0/#input-descriptor-object](https://identity.foundation/presentation-exchange/spec/v2.0.0/#input-descriptor-object)
+#[derive(Clone, Debug, Default, Serialize, Deserialize, PartialEq, Eq)]
+pub struct Constraints {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    fields: Option<Vec<ConstraintsField>>,
```
cobward: As above, may not be important to distinguish between `None` and an empty vec.

src/core/input_descriptor.rs
```diff
@@ -0,0 +1,610 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    name: Option<String>,
+    // Optional predicate value
+    predicate: Option<Predicate>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    filter: Option<serde_json::Value>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    optional: Option<bool>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    intent_to_retain: Option<bool>,
```
cobward: As above, may be better to use the default boolean value rather than `None`

src/core/presentation_definition.rs
```diff
@@ -0,0 +1,382 @@
...
+    input_descriptors: Vec<InputDescriptor>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    submission_requirements: Option<Vec<SubmissionRequirement>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    name: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    purpose: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    format: Option<ClaimFormatMap>,
+}
```
cobward: As above, can use `Vec<ClaimFormat>`

src/core/response/mod.rs
```diff
@@ -35,6 +38,51 @@ impl AuthorizationResponse {
...
+                // Ensure the definition id matches the submission's definition id.
+                if presentation_submission.definition_id() != presentation_definition.id() {
+                    bail!("Presentation Definition ID does not match the Presentation Submission.")
+                }
+
+                // Validate the VP Token as an unencoded payload.
+                response.vp_token().validate_unencoded(
+                    presentation_definition,
+                    presentation_submission.descriptor_map(),
+                )?;
```
cobward: This assumes that the vp_token is a W3C VCDM Verifiable Presentation. That is only one possibility. The VP token could be a base64-encoded `DocumentResponse` from ISO 18013-5/-7, an SD-JWT+KB, or a completely proprietary data format.

Validation is incredibly complex and dependent on many different contextual factors, so for now, I suggest removing this function in favour of individual methods implementing specific checks that we know we can always perform (such as the definition ID matching).

cobward (changes requested):

src/core/response/parameters.rs
```diff
@@ -46,28 +52,96 @@ impl From<VpToken> for Json {
...
+            Err(_) => Ok(serde_json::from_str(&self.0)?),
+        }
+    }
+
+    /// Return the Verifiable Presentation Token as a JSON object from a base64
+    /// encoded string.
+    pub fn decode_base64(&self) -> Result<Json, Error> {
+        let decoded = BASE64_STANDARD.decode(&self.0)?;
+        Ok(serde_json::from_slice(&decoded)?)
+    }
```
cobward: These methods are inaccurate:

> JSON String or JSON object that MUST contain a single Verifiable Presentation or an array of JSON Strings and JSON objects each of them containing a Verifiable Presentations. Each Verifiable Presentation MUST be represented as a JSON string (that is a Base64url encoded value) or a JSON object depending on a format as defined in Annex E of [[OpenID.VCI](https://openid.net/specs/openid-4-verifiable-presentations-1_0-20.html#OpenID.VCI)]

We have a bug in our definition of `VpToken`. It should be something like:

```rust
enum VpToken {
    Single(String),
    SingleAsMap(Map<String, Value>),
    Many(Vec<VpToken>),
}
```

src/core/response/parameters.rs
```diff
@@ -46,28 +52,96 @@ impl From<VpToken> for Json {
...
+    /// This method will return `Ok(())` if the VP token is valid.
+    ///
+    /// # Errors
+    ///
+    /// This method will return an error if the VP token is invalid,
+    /// or if the verifiable presentation is invalid or does not meet
+    /// the requirements of the presentation definition.
+    ///
+    ///
+    pub fn validate_unencoded(
```
cobward: As above, this is making an assumption that the VpToken is a VCDM VP. There's also some confusion here about encoded as a JWT vs unencoded. Encoded as a JWT refers to the authorization response (see [direct_post.jwt](https://openid.net/specs/openid-4-verifiable-presentations-1_0-20.html#name-response-mode-direct_postjw)), not the VP token. The VP token is never encoded directly as a JWT. It's either a String, a JSON object, or an array of VP tokens.

src/holder/verifiable_presentation_builder.rs
cobward, line 1: I think we said this should be removed from this library and added to SSI?

cobward (changes requested):

src/core/credential_format/mod.rs
Ryanmtate: The presentation exchange presentation definition test suite test case `format_example` shows `format` property as a map. I tried with the `Vec<ClaimFormat>` but encountered errors in the test case.

```json
{
  "presentation_definition": {
    "id": "32f54163-7166-48f1-93d8-ff217bdb0653",
    "input_descriptors": [],
    "format": {
      "jwt": {
        "alg": ["EdDSA", "ES256K", "ES384"]
      },
      "jwt_vc": {
        "alg": ["ES256K", "ES384"]
      },
      "jwt_vp": {
        "alg": ["EdDSA", "ES256K"]
      },
      "ldp_vc": {
        "proof_type": [
          "JsonWebSignature2020",
          "Ed25519Signature2018",
          "EcdsaSecp256k1Signature2019",
          "RsaSignature2018"
        ]
      },
      "ldp_vp": {
        "proof_type": ["Ed25519Signature2018"]
      },
      "ldp": {
        "proof_type": ["RsaSignature2018"]
      }
    }
  }
}

```

src/core/presentation_definition.rs
Ryanmtate: Removed the `Option` and but maintained `ClaimFormatMap` to pass test suite.

src/core/response/mod.rs
Ryanmtate: Validation methods have been removed in commit [0e0f9](https://github.com/spruceid/oid4vp-rs/pull/8/commits/0e0f985da4e5e0827efedf86509a424f25cc4816).

src/core/response/parameters.rs
Ryanmtate: I have a TODO in the code to update the VpToken. Maybe we can do this in another PR.

src/core/response/parameters.rs
Ryanmtate: This method has been removed.

src/holder/verifiable_presentation_builder.rs
Ryanmtate: Okay to remove in another PR once added to SSI?

cobward (approved):

src/holder/verifiable_presentation_builder.rs
```diff
@@ -0,0 +1,164 @@
+use std::time::{SystemTime, UNIX_EPOCH};
+
+use serde::{Deserialize, Serialize};
+use ssi_claims::jwt::{VerifiableCredential, VerifiablePresentation};
```
timothee-haudebourg, line 4: I think there is a confusion here, probably because of the lack of documentation in ssi I'll admit. Those two types are wrappers around the [JWT registered claims `vc` and `vp`](https://www.iana.org/assignments/jwt/jwt.xhtml) to be used with `ssi_claims::jwt::ClaimSet`.

If you want to represent VCs and VPs, the appropriate types are defined in `ssi_claims::vc`.

src/holder/verifiable_presentation_builder.rs
```diff
@@ -0,0 +1,164 @@
...
+use ssi_claims::vc::v2::syntax::VERIFIABLE_PRESENTATION_TYPE;
+use ssi_dids::ssi_json_ld::CREDENTIALS_V1_CONTEXT;
+use ssi_dids::{
+    ssi_json_ld::syntax::{Object, Value},
+    DIDURLBuf,
+};
+
+#[derive(Debug, Clone)]
+pub struct VerifiablePresentationBuilderOptions {
+    pub issuer: DIDURLBuf,
```
timothee-haudebourg, line 14: Is it a requirement of OID4VP that we use DIDs? I know this is not a requirement of VPs.

src/holder/verifiable_presentation_builder.rs
```diff
@@ -0,0 +1,164 @@
...
+    /// Returns an empty verifiable presentation builder.
+    pub fn new() -> Self {
+        Self(VerifiablePresentation(Value::Object(Object::new())))
+    }
+
+    /// Returns a verifiable presentation builder from options.
+    ///
+    /// This will set the issuance date to the current time and the expiration
+    /// date to the expiration secs from the issuance date.
+    pub fn from_options(options: VerifiablePresentationBuilderOptions) -> VerifiablePresentation {
```
timothee-haudebourg, line 51: This sounds like a re-implementation of VCDM1.0 JWT VP encoding. This is already implemented in `ssi` [here](https://github.com/spruceid/ssi/blob/bfdf22d14ac96bef7e8fc0c88fa9980198da699d/crates/claims/crates/vc/src/v1/jwt/mod.rs#L88).

```rust
// Create the presentation.
let presentation: ssi::claims::vc::v1::JsonPresentation = ...;

// Encode the JWT claims according to VCDM1.0 JWT VP format.
let jwt_claims = presentation.to_jwt_claims();

// Sign the claims into a JWT.
let jwt = jwt_claims.sign(signer).await.unwrap();
```

src/core/credential_format/mod.rs
```diff
@@ -1,17 +1,305 @@
...
+    /// specification. Expression of supported algorithms in relation to these formats MUST be conveyed using an JWT alg
+    /// property paired with values that are identifiers from the JSON Web Algorithms registry in
+    /// [RFC7518](https://identity.foundation/claim-format-registry/#ref:RFC7518) Section 3.
+    #[serde(rename = "jwt_vc")]
+    JwtVc,
+    /// See [JwtVc](JwtVc) for more information.
+    #[serde(rename = "jwt_vp")]
+    JwtVp,
+    #[serde(rename = "jwt_vc_json")]
+    JwtVcJson,
```
timothee-haudebourg, line 179: What is this format?

timothee-haudebourg (commented):
Adding a late review of this to clarify some stuff about `ssi`.

PR #18, opened 2024-09-18
update vp token definition

Ryanmtate:
This PR introduces updates to the `VpToken` definition:

```rust
pub enum VpToken {
    Single(Vec<u8>),
    SingleAsMap(Json),
    Many(Vec<VpToken>),
}
```

~~The PR includes a `TryFrom<VerifiablePresentation> for VpToken` method for convenience.~~

Includes a `TryFrom<AnyJsonPresentation> for VpToken` as a convenience method.


src/core/response/parameters.rs
```diff
@@ -26,18 +28,26 @@ impl From<IdToken> for Json {
...
+/// > format in the Credential Response, this rules MUST also be followed when encoding Credentials of
+/// > this format in the vp_token response parameter. Otherwise, this specification does not require
+/// > any additional encoding when a Credential format is already represented as a JSON object or a JSON string.
+///
+/// See: [OpenID.VP#section-6.1-2.2](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-6.1-2.2)
 #[derive(Debug, Clone)]
-pub struct VpToken(pub String);
+pub enum VpToken {
+    Single(String),
+    SingleAsMap(Json),
```
cobward: Better to represent as a `Map<String, Json>` to avoid having to unwrap the `Json::Object`

src/core/response/parameters.rs
```diff
@@ -47,13 +57,80 @@ impl TryFrom<Json> for VpToken {
     type Error = Error;
 
     fn try_from(value: Json) -> Result<Self, Self::Error> {
-        serde_json::from_value(value).map(Self).map_err(Into::into)
+        match value {
+            // NOTE: When parsing a Json string object, it must be base64 encoded.
+            Json::String(s) => Ok(Self::Single(BASE64_STANDARD.encode(s.as_bytes()))),
```
cobward: The spec says the value must be represented as a base64url JSON string, so you should be doing *base64url* instead of *base64*, and _decoding_ instead of _encoding_

src/core/response/parameters.rs
```diff
@@ -47,13 +57,80 @@ impl TryFrom<Json> for VpToken {
...
-    fn from(value: VpToken) -> Self {
-        value.0.into()
+impl TryFrom<VpToken> for Json {
+    type Error = Error;
+
+    fn try_from(value: VpToken) -> Result<Self, Self::Error> {
+        match value {
+            VpToken::Single(s) => {
+                // Decode base64 string.
+                let bytes = BASE64_STANDARD.decode(s.as_bytes())?;
```
cobward: As above, this should be base64url encoding

src/core/response/parameters.rs
```diff
@@ -26,18 +28,26 @@ impl From<IdToken> for Json {
...
+/// > If Appendix A of [OpenID.VCI] defines a rule for encoding the respective Credential
+/// > format in the Credential Response, this rules MUST also be followed when encoding Credentials of
+/// > this format in the vp_token response parameter. Otherwise, this specification does not require
+/// > any additional encoding when a Credential format is already represented as a JSON object or a JSON string.
+///
+/// See: [OpenID.VP#section-6.1-2.2](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-6.1-2.2)
 #[derive(Debug, Clone)]
-pub struct VpToken(pub String);
+pub enum VpToken {
+    Single(String),
```
cobward: Based on the above comments, this should be a `Vec<u8>` instead of String, as the decoded value may not be UTF-8.

src/core/response/parameters.rs
```diff
@@ -47,13 +57,80 @@ impl TryFrom<Json> for VpToken {
...
+impl TryFrom<VerifiablePresentation> for VpToken {
+    type Error = Error;
+
+    fn try_from(vp: VerifiablePresentation) -> Result<Self, Self::Error> {
+        Self::try_from(vp.0.into_serde_json())
+    }
+}
+
+impl TryFrom<VpToken> for Vec<VerifiablePresentation> {
+    type Error = Error;
```
cobward: These are incorrect. `ssi::claims::jwt::VerifiablePresentation` cannot be converted into a VP token. It's the JWT claim for a `VerifiablePresentation` that you use if you are encoding a VP as a JWT. In reality the `vp_token` would be a JWT (that has this `VerifiablePresentation` claim), a [JsonPresentation](https://docs.rs/ssi-vc/latest/ssi_vc/syntax/enum.AnyJsonPresentation.html), or something else (for example ISO 18013-7 uses a CBOR encoded binary format for vp_tokens)

cobward (changes requested):

src/core/response/parameters.rs
Ryanmtate: > Each Verifiable Presentation MUST be represented as a JSON string (that is a Base64url encoded value)


Wouldn't this be encoded and not decoded? Is the inner `String` value not what is being sent over the wire?

src/core/response/parameters.rs
Ryanmtate: Yeah, I wondered about using `JsonPresentation` type after reviewing Tim's SSI PR.

I'm going off of the https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-6.1-2.2 definition. Is there a better definition to review. I didn't see anything about CBOR encoding.

src/core/response/parameters.rs
Ryanmtate: Commit [ee67e6b](https://github.com/spruceid/oid4vp-rs/pull/18/commits/ee67e6bb988c505ce7a73708fa0189975bd021a3) changes to `BASE64_URL_SAFE_NO_PAD`

src/core/response/parameters.rs
cobward: What is sent over the wire is `JSON::String(Base64UrlEncode(<Value>))`, for example the authorization response could be:

```
 HTTP/1.1 302 Found
  Location: https://client.example.org/cb#
    presentation_submission=...
    &vp_token=%22ZXlKaGJHY2lPaUpJVXpJMU5pSXNJblI1Y0NJNklrcFhWQ0o5LmV5SnpkV0lpT2lJeE1qTTBOVFkzT0Rrd0lpd2libUZ0WlNJNklrcHZhRzRnUkc5bElpd2lhV0YwSWpveE5URTJNak01TURJeWZRLlNmbEt4d1JKU01lS0tGMlFUNGZ3cE1lSmYzNlBPazZ5SlZfYWRRc3N3NWM%22
```

Ryanmtate: Is the right way to think about this to include a method on the `VpToken` type itself for encoding the payload for the various variants. E.g.,

```rust
impl VpToken {
    pub fn as_base64_url_encoded(self) -> Result<String, Error> {
        let json: Json = self.try_into()?;
        let string = serde_json::to_string(&json)?;
        Ok(BASE64_URL_SAFE_NO_PAD.encode(string.as_bytes()))
    }
}
```

src/core/response/parameters.rs
Ryanmtate: Commit [ae2c5c9](https://github.com/spruceid/oid4vp-rs/pull/18/commits/ae2c5c983fb15c4910a85c309498245a7d30bbfa) replaces the `VerifiablePresentation` with `AnyJsonPresentation`. It also removes the `VerifiablePresentatioBuilder` in favor of `JsonPresentation`

src/core/response/parameters.rs
Ryanmtate: > What is sent over the wire is `JSON::String(Base64UrlEncode(<Value>))`, for example the authorization response could be:
> 
> ```
>  HTTP/1.1 302 Found
>   Location: https://client.example.org/cb#
>     presentation_submission=...
>     &vp_token=%22ZXlKaGJHY2lPaUpJVXpJMU5pSXNJblI1Y0NJNklrcFhWQ0o5LmV5SnpkV0lpT2lJeE1qTTBOVFkzT0Rrd0lpd2libUZ0WlNJNklrcHZhRzRnUkc5bElpd2lhV0YwSWpveE5URTJNak01TURJeWZRLlNmbEt4d1JKU01lS0tGMlFUNGZ3cE1lSmYzNlBPazZ5SlZfYWRRc3N3NWM%22
> ```

Based on your comment above, is the following statement still accurate?

> Based on the above comments, this should be a Vec<u8> instead of String, as the decoded value may not be UTF-8.

Seems to me that `VpToken::Single` would contain `JSON::String(Base64UrlEncode(<Value>))`, which would be a `serde_json::Value` type and not a `Vec<u8>`.

Confused by your comments.

src/core/response/parameters.rs
```diff
@@ -47,13 +57,48 @@ impl TryFrom<Json> for VpToken {
...
+            // of the VP token.
+            Json::String(s) => Ok(Self::Single(BASE64_URL_SAFE_NO_PAD.decode(s)?)),
+            // NOTE: When the Json is an object, it must be a map.
+            Json::Object(map) => Ok(Self::SingleAsMap(map)),
+            Json::Array(arr) => {
+                let mut tokens = Vec::new();
+                for value in arr {
+                    tokens.push(Self::try_from(value)?);
+                }
+                Ok(Self::Many(tokens))
```
cobward: Style comment: we prefer iterators over for-loops pushing into mutable vecs:
```suggestion
                arr.into_iter()
                    .map(Self::try_from)
                    .collect::<Result<Vec<Self>, Self::Error>>()
                    .map(Self::Many)
```

src/core/response/parameters.rs
```diff
@@ -47,13 +57,48 @@ impl TryFrom<Json> for VpToken {
...
+            VpToken::Many(tokens) => {
+                let mut arr: Vec<Json> = Vec::new();
+                for token in tokens {
+                    arr.push(token.try_into()?);
+                }
+                Ok(arr.into())
+            }
+        }
+    }
+}
```
cobward: This can be rewritten as `From<VpToken> for Json`:
```suggestion
impl From<VpToken> for Json {
    fn from(value: VpToken) -> Self {
        match value {
            VpToken::Single(s) => serde_json::Value::String(BASE64_URL_SAFE_NO_PAD.encode(s)),
            VpToken::SingleAsMap(map) => serde_json::Value::Object(map),
            VpToken::Many(tokens) => Self::Array(tokens.into_iter().map(Self::from).collect()),
        }
    }
}
```

src/core/response/parameters.rs
```diff
@@ -47,13 +57,48 @@ impl TryFrom<Json> for VpToken {
...
+
+impl TryFrom<AnyJsonPresentation> for VpToken {
+    type Error = Error;
+
+    fn try_from(vp: AnyJsonPresentation) -> Result<Self, Self::Error> {
+        Ok(VpToken::Single(serde_json::to_vec(&vp)?))
```
cobward: A `JsonPresentation` is a W3C Verifiable Presentation, which is a JSON object. Therefore it should be represented as `VpToken::SingleAsMap` ([example in the OID4VP spec](https://openid.net/specs/openid-4-verifiable-presentations-1_0-20.html#appendix-A.1.2.3-5)).

cobward (changes requested):

cobward (approved):

tests/jwt_vp.rs
```diff
@@ -21,28 +16,51 @@ pub async fn create_test_verifiable_presentation() -> Result<String> {
     .unwrap();
 
     let holder_did = DIDKey::generate_url(signer.jwk())?;
-    let verifier_did = DIDKey::generate_url(&verifier)?;
+
+    println!("holder_did: {:?}", holder_did);
```
w4ll3: should this print be here?

PR #19, opened 2024-09-19
add credential_type method to constraint field

Ryanmtate:
This PR adds a helper method to parse the credential type from a constraint field.

### DIFF
--- a/src/core/input_descriptor.rs
+++ b/src/core/input_descriptor.rs
@@ -1,3 +1,5 @@
+use std::collections::HashSet;
+
 use super::{credential_format::*, presentation_submission::*};
 use crate::utils::NonEmptyVec;
 
@@ -135,6 +137,23 @@ impl InputDescriptor {
         self
     }
 
+    /// Return the format of the input descriptor.
+    ///
+    /// The Input Descriptor Object MAY contain a format property. If present,
+    /// its value MUST be an object with one or more properties matching the registered
+    /// Claim Format Designations (e.g., jwt, jwt_vc, jwt_vp, etc.).
+    ///
+    /// This format property is identical in value signature to the top-level format object,
+    /// but can be used to specifically constrain submission of a single input to a subset of formats or algorithms.
+    pub fn format(&self) -> &ClaimFormatMap {
+        &self.format
+    }
+
+    /// Return the format designations of the input descriptor as a hash set.
+    pub fn format_designations(&self) -> HashSet<&ClaimFormatDesignation> {
+        self.format.keys().collect()
+    }
+
     /// Set the group of the constraints field.
     pub fn set_group(mut self, group: Vec<GroupId>) -> Self {
         self.group = group;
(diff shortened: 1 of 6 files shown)


src/core/input_descriptor.rs
```diff
@@ -588,6 +595,53 @@ impl ConstraintsField {
             })
             .collect()
     }
+
+    /// Return the Credential Type of the constraints field
+    pub fn credential_type(&self) -> Option<CredentialType> {
```
cobward: Writing this type of function seems like a losing battle. Should we instead just apply the PresentationDefinition to each credential and see which ones match?

Alternatively I would be ok with calling this function `credential_type_hint`, as we probably can't exhaustively list the CredentialTypes that are being requested.

cobward: Could there be more than one type? If so, should this function return `Vec<CredentialType>`?

Ryanmtate: There is a case where a constraint field can have a `pattern` filter, which could include multiple credentials to parse on. 

I've updated the return type to a `Vec<CredentialType>` in that case. 

[521895e](https://github.com/spruceid/oid4vp-rs/pull/19/commits/521895eb53942beee1cafcccd050ea992b3d02cd) provides a unit test and example presentation definition for this case.

It should be noted, however, that the input descriptor can have multiple constraint fields, in which case, there would be a collection of credential types found in the input descriptor, and then again, multiple found amongst all input descriptors.

Ryanmtate: > Writing this type of function seems like a losing battle. Should we instead just apply the PresentationDefinition to each credential and see which ones match?

Not sure I understand this comment.

It seemed like the best way to parse the credential types requested from a presentation definition were from its input descriptors, which include constraint fields that would define the path to the `$.vc.type` or `$.type` field that would denote the credential requested for information.

Is there a better way to parse a presentation definition?

To your point about `hint`, it is not guaranteed that the presentation definition will include a `$.type` constraint field. 

Indeed, we would need to check which credentials include the information requested in the input descriptor. This is a more expensive reverse lookup through the VDC collection to deserialize/decode credentials to find a match.

Currently, this is a [TODO](https://github.com/spruceid/mobile-sdk-rs/blob/skit-459-integrate-oid4vp-into-mobile-sdk-rust-swift-kotlin/src/oid4vp/holder.rs#L139-L148) in `mobile-sdk-rs` that relates to these questions.

I've updated the function name to your `credential_type_hint` suggestion and made a note about it potentially being empty in the code comments.

This method (or something like it) will help to simplify the credential retrieval from the VDC collection in mobile-sdk-rs.

cobward: > Not sure I understand this comment.
> 
> It seemed like the best way to parse the credential types requested from a presentation definition were from its input descriptors, which include constraint fields that would define the path to the $.vc.type or $.type field that would denote the credential requested for information.
> 
> Is there a better way to parse a presentation definition?

Sorry for not being clearer. What I meant was that "parsing a presentation definition to extract constraints for matching against a set of credentials" is a losing battle. Doing anything other than "apply the JSON schemas to each credential in the set and see which ones match" is either reimplementing the entirety of the JSON schema processing code or is only handling special cases. Most wallets will only have a handful of credentials, so loading them up one by one to validate the JSON schema is not going to be a slow process.

cobward: A concrete example of this not working, for mdocs the JSON path looks like this:
`$['org.iso.18013.5.1']['birth_date']`

Where the first path element is identifying the document type (AKA credential type) `org.iso.18013.5.1`.

Ryanmtate: I've added a method `Credential::check_presentation_definition` in mobile-sdk-rs which will check the credential against a presentation definition reference. See: https://github.com/spruceid/mobile-sdk-rs/blob/skit-459-integrate-oid4vp-into-mobile-sdk-rust-swift-kotlin/src/credential/mod.rs#L111-L179

This check will use the type hint, if it exists, otherwise will default to using the JSON schema to parse the JSON-encoded credential against the input descriptor constraint fields.

I still think the type hint has some value here, but I agree it is limited and a full credential check is necessary.

The `check_presentation_definition` method is currently on `Credential`, but may be moved to `ParsedCredential`. Will explore this in a subsequent PR in mobile-sdk-rs. For example, still needed in a future PR is the handling of an Mdoc type, which might use `Element` type for checking its value.

PR #20, opened 2024-09-20
refactor human readable string utils

Ryanmtate:
Cleans up the `requested_fields` method on the constraints field and moves the human readable string utility to the `utils` file.

### DIFF
--- a/src/core/input_descriptor.rs
+++ b/src/core/input_descriptor.rs
@@ -1,7 +1,7 @@
 use std::collections::HashSet;
 
 use super::{credential_format::*, presentation_submission::*};
-use crate::utils::NonEmptyVec;
+use crate::utils::{to_human_readable_string, NonEmptyVec};
 
 use anyhow::{bail, Context, Result};
 use jsonschema::{JSONSchema, ValidationError};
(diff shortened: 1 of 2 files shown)


src/core/input_descriptor.rs
```diff
@@ -599,17 +599,13 @@ impl ConstraintsField {
...
-    /// This will also stripe the periods from the JSON path and return the last word in the path.
-    ///
-    /// e.g., `["$.verifiableCredential.credentialSubject.dateOfBirth"]` will return `["Date of Birth"]`.
-    /// e.g., `["$.verifiableCredential.credentialSubject.familyName"]` will return `["Family Name"]`.
+    /// This will stripe the delimiters from the JSON path and return the last value in the path.
     ///
+    /// e.g., `["$.verifiableCredential.credentialSubject.dateOfBirth"]` will return `["dateOfBirth"]`.
+    /// e.g., `["$.verifiableCredential.credentialSubject.familyName"]` will return `["familyName"]`.
     pub fn requested_fields(&self) -> Vec<String> {
         self.path()
```
cobward, line 610: I'm actually wondering if we should just get rid of this function, as it only supports a very specific subset of JSON Path expressions. How would it handle inputs like:
```
$['verifiableCredential']['credentialSubject']['achievements'][*]['id']
$['verifiableCredential']['credentialSubject']['capabilities'][*]
$['verifiableCredential']['credentialSubject']['achievements'][?(@.evidence)]
```

Take a look at the `Example JSONPath Expressions` table [here](https://identity.foundation/presentation-exchange/spec/v2.0.0/#jsonpath-syntax-definition) for more examples.

Ryanmtate: Yeah, I agree. We may wish to move this method closer to the parsed credential that matches the input descriptor.

After reviewing the JSONPath expressions, I don't think it belongs here either.

Ryanmtate:
Closing this for now. Can revive if we feel different.

Goal is to move this functionality closer to the parsed credential.

We should avoid deleting the branch until further notice, as we may wish to refer to some utility methods.

PR #22, opened 2024-09-27
Add jsonpath as a re-export under deps module

Ryanmtate:


src/lib.rs
```diff
@@ -4,3 +4,7 @@ pub(crate) mod tests;
 mod utils;
 pub mod verifier;
 pub mod wallet;
+
+pub mod deps {
+    pub use jsonpath_lib;
+}
```
cobward: It's more common to just re-export without a deps module in Rust, as the generated documentation identifies them as re-exports ([example](https://docs.rs/p256/latest/p256/#reexports))
```suggestion
pub use jsonpath_lib;
```

PR #25, opened 2024-10-02
Add SdJwtVc claim format as a serde_json::Value type

Joey-Silberman:
This PR adds `SdJwtVc` as a non-normative claim format. It's not a normative claim format because it's not specified in the claim format registry. Because `SdJwtVc` is not listed in the [Decentralized Identity presentation definition claim format designation](https://github.com/decentralized-identity/claim-format-registry/blob/main/schemas/presentation-definition-claim-format-designations.json), it makes sense to keep it as non-normative to avoid restricting the format.

### DIFF
--- a/src/core/credential_format/mod.rs
+++ b/src/core/credential_format/mod.rs
@@ -77,6 +77,8 @@ pub enum ClaimFormat {
     },
     #[serde(rename = "mso_mdoc")]
     MsoMDoc(serde_json::Value),
+    #[serde(rename = "sd_jwt_vc")]
+    SdJwtVc(serde_json::Value),
     /// Support for non-standard claim formats.
     // NOTE: a `format` property will be included within the serialized
     // type. This will help for identifying the claim format designation type.
@@ -101,6 +103,7 @@ impl ClaimFormat {
             ClaimFormat::AcVc { .. } => ClaimFormatDesignation::AcVc,
             ClaimFormat::AcVp { .. } => ClaimFormatDesignation::AcVp,
             ClaimFormat::MsoMDoc(_) => ClaimFormatDesignation::MsoMDoc,
+            ClaimFormat::SdJwtVc(_) => ClaimFormatDesignation::SdJwtVc,
             ClaimFormat::Other(value) => {
                 // Parse the format from the first key found in the value map.
                 let format = value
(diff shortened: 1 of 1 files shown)


src/core/credential_format/mod.rs
```diff
@@ -213,6 +216,10 @@ pub enum ClaimFormatDesignation {
     /// the Credential format can be utilized with any type of Credential (or mdoc document types).
     #[serde(rename = "mso_mdoc")]
     MsoMDoc,
+    /// The format is defined by IETF Web Authorization Protocol (https://datatracker.ietf.org/doc/draft-ietf-oauth-selective-disclosure-jwt/)
+    /// which defines a Selective Disclosure for JWTs (SD-JWT)
```
timothee-haudebourg, line 220: The description should be more precise about the credential format. SD-JWT is just the enveloppe, and that doesn't say anything about how the credential is encoded as the enveloppe's payload (technically we just know it's JSON). What we want to target is the [VCDM2.0 JOSE COSE SD-JWT](https://w3c.github.io/vc-jose-cose/#with-sd-jwt) format. It's a VCDM2.0 credential, interpreted as JWT claims, envelopped into an SD-JWT.

src/core/credential_format/mod.rs
```diff
@@ -213,6 +216,10 @@ pub enum ClaimFormatDesignation {
...
+    #[serde(rename = "sd_jwt_vc")]
+    SdJwtVc,
```
timothee-haudebourg, line 222: I suggested using `spruceid_vcdm2_sd_jwt` as a name for this variant (`spruceid_` to make it clear it's a custom extension). I'll admit this might have been a bit excessive :sweat_smile: but with this name I'm affraid of a collision with an eventual future official addition to the DIF registry, and also that the `jwt_vc` part could be conflated with the legacy ["JWT-VC" format from the VCDM1.1 spec](https://www.w3.org/TR/vc-data-model/#json-web-token). What do you think @cobward?

src/core/credential_format/mod.rs
```diff
@@ -77,6 +77,8 @@ pub enum ClaimFormat {
     },
     #[serde(rename = "mso_mdoc")]
     MsoMDoc(serde_json::Value),
+    #[serde(rename = "sd_jwt_vc")]
+    SdJwtVc(serde_json::Value),
```
timothee-haudebourg, line 81: What is this `serde_json::Value`?

timothee-haudebourg (changes requested):

src/core/credential_format/mod.rs
cobward: Agreed, let's err on the side of over-specifying. I would also be open to not including this in the OID4VP library. We can just use the `Other` variants since this isn't a standard value.

timothee-haudebourg: I think you're right, I'm in favor of just using the `Other` variant and delegate the implementation of the VCDM2.0 JOSE COSE SD-JWT profile as an application specific extension, implemented in the application.

Joey-Silberman:
Closing this PR because it turns out we won't be adding a custom type directly in oid4vp-rs since there is already the Other variant that can allow us to implement that as a custom extension in the mobile-sdk directly.

PR #24, opened 2024-10-01
Skit 563 fix

Ryanmtate:
https://github.com/spruceid/oid4vp-rs/pull/19 merged into the wrong branch. 

Pushing this branch to fix the incorrect merge.

PR #26, opened 2024-10-04
add associative method to presentation definition

Ryanmtate:
This PR moves the `check_presentation_definition` method in `mobile-sdk-rs` to an associative method on `PresentationDefinition`, which is supplied a `serde_json::Value` encoded credential to check the constraint fields against.

### DIFF
--- a/src/core/input_descriptor.rs
+++ b/src/core/input_descriptor.rs
@@ -1,7 +1,8 @@
-use std::collections::HashSet;
-
 use super::{credential_format::*, presentation_submission::*};
 use crate::utils::NonEmptyVec;
+
+use std::collections::HashSet;
+use std::sync::Arc;
 
 use anyhow::{bail, Context, Result};
 use jsonschema::{JSONSchema, ValidationError};
@@ -235,7 +236,7 @@ impl InputDescriptor {
                 found_elements = true;
 
                 // If a filter is available with a valid schema, handle the field validation.
-                if let Some(Ok(schema_validator)) = validator.as_ref() {
+                if let Some(schema_validator) = validator.as_ref() {
                     let validated_fields = field_elements.iter().find(|element| {
                         match schema_validator.validate(element) {
                             Err(errors) => {
@@ -398,6 +399,75 @@ impl Constraints {
     }
 }
 
+/// This type is non-normative and used to simplify the construction
+/// of constraint field validators at deserialization and construction,
+/// thereby reducing runtime error checking requirements.
+///
+/// Only the inner raw JSON value is serialized and deserialized.
+#[derive(Debug, Clone)]
+pub struct ConstraintsFieldValidator {
+    raw: serde_json::Value,
+    compiled: Arc<JSONSchema>,
+}
+
+impl ConstraintsFieldValidator {
+    pub fn validator(&self) -> &Arc<JSONSchema> {
+        &self.compiled
+    }
+}
+
+impl AsRef<serde_json::Value> for ConstraintsFieldValidator {
+    fn as_ref(&self) -> &serde_json::Value {
+        &self.raw
+    }
+}
+
+impl<'a> TryFrom<&'a serde_json::Value> for ConstraintsFieldValidator {
+    type Error = ValidationError<'a>;
+
+    fn try_from(value: &'a serde_json::Value) -> Result<Self, Self::Error> {
+        let compiled = JSONSchema::compile(value)?;
+        Ok(Self {
+            raw: value.to_owned(),
+            compiled: Arc::new(compiled),
+        })
+    }
+}
+
+// NOTE: implementing PartialEq directly due to JSONSchema not implementing PartialEq.
+impl PartialEq for ConstraintsFieldValidator {
+    fn eq(&self, other: &Self) -> bool {
+        self.raw == other.raw
+    }
+}
+
+impl Eq for ConstraintsFieldValidator {}
+
+impl Serialize for ConstraintsFieldValidator {
+    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        self.raw.serialize(serializer)
+    }
+}
+
+impl<'de> Deserialize<'de> for ConstraintsFieldValidator {
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+        D::Error: std::error::Error,
+    {
+        let raw = serde_json::Value::deserialize(deserializer)?;
+
+        let compiled = JSONSchema::compile(&raw).map(Arc::new).map_err(|e| {
+            serde::de::Error::custom(format!("Failed to compile JSON schema: {}", e))
+        })?;
+
+        Ok(ConstraintsFieldValidator { raw, compiled })
+    }
+}
+
 /// ConstraintsField objects are used to describe the constraints that a
 /// [Holder](https://identity.foundation/presentation-exchange/spec/v2.0.0/#term:holder)
 /// must satisfy to fulfill an Input Descriptor.
@@ -415,7 +485,7 @@ pub struct ConstraintsField {
     // Optional predicate value
     predicate: Option<Predicate>,
     #[serde(skip_serializing_if = "Option::is_none")]
-    filter: Option<serde_json::Value>,
+    filter: Option<ConstraintsFieldValidator>,
     #[serde(skip_serializing_if = "Option::is_none")]
     optional: Option<bool>,
     #[serde(default)]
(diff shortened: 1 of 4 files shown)


src/core/presentation_definition.rs
```diff
@@ -263,6 +264,50 @@ impl PresentationDefinition {
...
+
+                        // NOTE: if the field is not required and is not a match,
+                        // skip and continue checking the other fields.
+                        if !is_valid && field.is_optional() {
+                            return None;
+                        }
+
+                        Some(is_valid)
+                    }
+                    _ => None,
```
cobward: The error case here is not handled. What should this library do if it receives a JSONSchema filter that cannot be parsed? Based on the [specification](https://openid.net/specs/openid-4-verifiable-presentations-1_0-20.html#section-6.4-5.2) I believe we should outright reject the request. So I recommend changing the type of `ConstraintsField.filter` to `JSONSchema`, so that the failure happens during the parsing of the `PresentationDefinition`, and there is no error case at this point to handle.

src/core/presentation_definition.rs
```diff
@@ -263,6 +264,50 @@ impl PresentationDefinition {
...
+                })
+                .all(|is_valid| is_valid)
```
cobward: NABD: Can simplify by replacing `filter_map` with `all`:
```suggestion
                .all(|field| match field.validator() {
                    Some(Ok(validator)) => {
                        let is_valid = field
                            .path()
                            .iter()
                            .filter_map(|path| match selector_lock.try_write() {
                                Ok(mut selector) => selector(path).ok(),
                                _ => None,
                            })
                            .flatten()
                            // NOTE: This is currently assuming that if any of the paths are a match
                            // to the credential, then the validation is, at least partially, successful,
                            // and the credential may satisfy the presentation definition.
                            .any(|value| validator.validate(value).is_ok());

                        // NOTE: if the field is not required and is not a match,
                        // skip and continue checking the other fields.
                        if !is_valid && field.is_optional() {
                            return true;
                        }

                        is_valid
                    }
                    _ => true,
                })
```

src/core/presentation_definition.rs
```diff
@@ -263,6 +264,50 @@ impl PresentationDefinition {
...
+                .iter()
+                .filter_map(|field| match field.validator() {
+                    Some(Ok(validator)) => {
+                        let is_valid = field
+                            .path()
+                            .iter()
+                            .filter_map(|path| match selector_lock.try_write() {
+                                Ok(mut selector) => selector(path).ok(),
+                                _ => None,
+                            })
```
cobward: NABD: This is all happening serially on one thread of execution, so RwLock is not needed:
```suggestion
        let mut selector = jsonpath_lib::selector(credential);

        self.input_descriptors().iter().any(|descriptor| {
            descriptor
                .constraints()
                .fields()
                .iter()
                .filter_map(|field| match field.validator() {
                    Some(Ok(validator)) => {
                        let is_valid = field
                            .path()
                            .iter()
                            .filter_map(|path| selector(path).ok())
```

cobward (changes requested):
Also wondering if this should target the main branch? Are we still planning on merging skit-563-fix into main as it is?

src/core/presentation_definition.rs
```diff
@@ -263,6 +264,50 @@ impl PresentationDefinition {
...
+                            // NOTE: This is currently assuming that if any of the paths are a match
+                            // to the credential, then the validation is, at least partially, successful,
+                            // and the credential may satisfy the presentation definition.
+                            .any(|value| validator.validate(value).is_ok());
+
+                        // NOTE: if the field is not required and is not a match,
+                        // skip and continue checking the other fields.
+                        if !is_valid && field.is_optional() {
+                            return None;
+                        }
```
cobward: To save a bit of computation you could move this `is_optional` check to before the constraints validation.

src/core/presentation_definition.rs
Ryanmtate: Given the mutability of `selector`, needed something here the offers inner mutability for the selector.

src/core/presentation_definition.rs
Ryanmtate: I agree with the recommendation to change `filter` type to `JSONSchema`. 

The downside is `JSONSchema` does not implement `serde` traits, which causes some grief for serialization with the `ConstraintsField`. 

For now, I am skipping the serialization/deserialization for the `validator` field, which means that it will show up as `None`, and will require an additional check. 

It would be nice to move the validation error earlier in the construction of the definition, versus runtime.

I've added a check on the `set_filter` method to perform this check on construction.

I've kept the original `validator()` method, but renamed it `try_validator`:

```rust
pub fn try_validator(&self) -> Option<Result<JSONSchema, ValidationError>> {
        self.filter.as_ref().map(JSONSchema::compile)
    }
```

But, I've changed the `validator` method to return the compiled JSONSchema as an Option if it is exists:

```rust
pub fn validator(&self) -> Option<Arc<JSONSchema>> {
        self.validator.clone()
    }
```

In the interest of time and as a temporary solution, I'll change `check_credential_validation` to return a result type and use the `try_validator` method at runtime.

The longer-term solution, which I'd prefer to come in a subsequent PR would be to write a custom serializer/deserializer to handle the `JSONSchema` type, or even the constraints field itself.

Ryanmtate: See: [0ddcdea](https://github.com/spruceid/oid4vp-rs/pull/26/commits/0ddcdea06db7846c1d32e65a94f6888a69770f37)

src/core/presentation_definition.rs
Ryanmtate: On further inspection, I think the RwLock was only needed previously where I was implementing this method, but since moving into the presentation definition scope, removing the RwLock is successful. See [24395df](https://github.com/spruceid/oid4vp-rs/pull/26/commits/24395df27994c0ed120d6bb6d5ea6ca269b2778c)

src/core/presentation_definition.rs
Ryanmtate: I've updated this flow to handle the validation error of the schema. See note: [344055c](https://github.com/spruceid/oid4vp-rs/pull/26/commits/344055c6dfea334c498a21f12639c032377d9e21)

In the future, we'll likely deprecate `try_validator` in favor of handling the compilation of JSONSchema on construction and deserialization.

This will require writing a custom deserializer for the `ConstraintField`, but I'd prefer to leave this to another PR.

Ryanmtate:
> Also wondering if this should target the main branch? Are we still planning on merging skit-563-fix into main as it is?

#24 Was actually a previous PR that was already accepted, but I see some new comments. I'll try to resolve those soon.

src/core/input_descriptor.rs
```diff
@@ -557,7 +585,19 @@ impl ConstraintsField {
     /// # Errors
```
Ryanmtate: Remove comments re Errors

src/core/input_descriptor.rs
```diff
@@ -416,12 +417,35 @@ pub struct ConstraintsField {
     predicate: Option<Predicate>,
     #[serde(skip_serializing_if = "Option::is_none")]
     filter: Option<serde_json::Value>,
+    // NOTE: Skip the JSON schema from serialization,
+    // and use the filter value for constructing the schema.
+    // NOTE: This will mean the validation will be None on
+    // deserialization, and will need to be set manually.
+    #[serde(skip)]
+    // NOTE: Wrapping JSONSchema in an Arc for cloning.
+    validator: Option<Arc<JSONSchema>>,
```
cobward: This is going to make it very difficult to construct a presentation definition on the verifier side. Instead, can you replace this with a new type?
```suggestion
    validator: ConstraintsFieldValidator
```

```rust
pub struct ConstraintsFieldValidator {
    raw: serde_json::Value,
    compiled: JSONSchema,
}

impl Serialize for ConstraintsFieldValidator {
    // Just serialize `self.raw`
}

impl Deserialize for ConstraintsFieldValidator {
    // Do `serde_json::Value::deserialize(d)` and then try to construct the JSONSchema.
}
```

src/core/presentation_definition.rs
```diff
@@ -263,6 +264,66 @@ impl PresentationDefinition {
...
+                            .flatten()
+                            // NOTE: This is currently assuming that if any of the paths are a match
+                            // to the credential, then the validation is, at least partially, successful,
+                            // and the credential may satisfy the presentation definition.
+                            .any(|value| validator.validate(value).is_ok());
+
+                        Ok(Some(is_valid))
+                    }
+                    // Handle the case where the constraint field does contain a valid filter validator.
+                    Some(Err(e)) => Err(e),
```
cobward: If you use the new type approach that I suggested above you don't need to handle an error here.

src/core/input_descriptor.rs
cobward: I see your comment was that you were concerned about time. I'll do up a commit for this change.

cobward (changes requested):

src/core/presentation_definition.rs
```diff
@@ -310,19 +297,11 @@ impl PresentationDefinition {
                             // and the credential may satisfy the presentation definition.
                             .any(|value| validator.validate(value).is_ok());
 
-                        Ok(Some(is_valid))
+                        is_valid
                     }
-                    // Handle the case where the constraint field does contain a valid filter validator.
-                    Some(Err(e)) => Err(e),
-                    _ => Ok(None),
+                    _ => false,
```
Ryanmtate: This might actually need to be `true` to handle `None` variant.

Ryanmtate: 020076a fixes.

cobward: Yep, should be true

src/core/metadata/mod.rs
```diff
@@ -84,6 +87,10 @@ impl WalletMetadata {
             ClaimFormatDesignation::JwtVcJson,
             ClaimFormatPayload::AlgValuesSupported(alg_values_supported.clone()),
         );
+        vp_formats_supported.insert(
```
cobward: This is the static wallet metadata as defined in the spec:
https://openid.net/specs/openid-4-verifiable-presentations-1_0-20.html#name-a-set-of-static-configurati

If you need different wallet metadata downstream then you will need to construct it.

cobward (changes requested):

src/core/metadata/mod.rs
Ryanmtate: Ah, okay, let me construct the metadata at instantiation.

cobward (approved):

PR #27, opened 2024-10-09
Fix/did client id scheme

Ryanmtate:


cobward:
As commented on the other PR, this is the static set of wallet metadata defined [here](https://openid.net/specs/openid-4-verifiable-presentations-1_0-20.html#name-a-set-of-static-configurati), so should not be changed.

PR #29, opened 2024-10-11
add method for adding client id scheme to wallet holder metadata

Ryanmtate:


src/core/metadata/mod.rs
```diff
@@ -47,6 +49,43 @@ impl WalletMetadata {
...
 
+    /// Return a mutable reference to the inner untyped object.
+    /// This is useful for adding additional parameters to the wallet metadata,
+    /// such as `client_id_schemes_supported`.
+    ///
+    /// NOTE: There is a typed method for adding a `client_id_schemes_supported` parameter.
+    /// See:
+    pub fn untyped_object_mut(&mut self) -> &mut UntypedObject {
+        &mut self.0
+    }
```
cobward: This is not needed, `DerefMut` is implemented for `WalletMetadata`, targeting `UntypedObject`.

src/core/metadata/mod.rs
```diff
@@ -47,6 +49,43 @@ impl WalletMetadata {
...
+            }
+            _ => ClientIdSchemesSupported::default(),
+        };
+
+        // Insert the scheme.
+        supported.0.push(client_id_scheme);
+
+        // Insert the updated client IDs schemes supported.
+        self.0.insert(supported);
+    }
```
cobward: I would prefer to not add this to the library. It's easy enough for an implementer to do `metadata.deref_mut()...` to access the UntypedObject and do this themselves. If we add this function then it begs the question of why don't we support setting, updating and mutating for all metadata parameters.

cobward: If you really want to keep this function, it can be tidied up a bit by using `get_or_default()` instead of `remove()`

src/core/metadata/mod.rs
```diff
@@ -47,6 +49,43 @@ impl WalletMetadata {
...
+            Some(Ok(supported)) => supported,
+            Some(Err(e)) => {
+                // NOTE: Not handling an error here, but warning if the client id schemes
+                // cannot be parsed. We may wish to handle this error in the future.
+                tracing::warn!(
+                    "Failed to parsed client id schemes supported from wallet metadata: {e:?}"
+                );
+
+                ClientIdSchemesSupported::default()
+            }
```
cobward: I don't think we should hide this error. This error will happen when the existing `client_id_schemes_supported` is malformed. I think it's unexpected for this method to overwrite the value rather than bubbling up that error.

cobward (changes requested):

src/core/metadata/mod.rs
Ryanmtate: I think it's okay to add methods as needed.

cobward (approved):

PR #33, opened 2024-10-15
parse requested fields from JSON encoded credential

Ryanmtate:
This PR provides functionality to parse the requested fields given a credential that the constraint field filter is applied to.

Additionally, `credential_type_hint` is removed in this PR in favor of using the constraint field filter.

### DIFF
--- a/src/core/input_descriptor.rs
+++ b/src/core/input_descriptor.rs
@@ -8,6 +8,7 @@ use anyhow::Result;
 use jsonschema::ValidationError;
 use serde::{Deserialize, Serialize};
 use serde_json_path::JsonPath;
+use uuid::Uuid;
 
 /// A GroupId represents a unique identifier for a group of Input Descriptors.
 ///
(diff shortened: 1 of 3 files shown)


src/core/input_descriptor.rs
```diff
@@ -834,3 +822,66 @@ pub enum ConstraintsLimitDisclosure {
...
+    pub fn constraint_field_id(&self) -> Option<String> {
+        self.constraint_field_id.clone()
+    }
+
+    /// Return the raw values of the credential matching the
+    /// constraint fields filter.
+    pub fn raw_fields(&self) -> Option<&serde_json::Value> {
+        self.raw_fields.as_ref()
+    }
+}
```
cobward: These fields are all public, so the getters and the constructor are not needed. In particular, constructing via `RequestedField { ... }` is preferred over a constructor, particular as it is easy in this case to get the parameters in the wrong order.
```suggestion
```

src/core/input_descriptor.rs
```diff
@@ -662,67 +660,57 @@ impl ConstraintsField {
...
+            // the first match that exists, if any.
+            //
+            // It may also be possible that multiple paths match,
+            // in which case there may be multiple raw fields
+            // that can be requested.
+            //
+            // As a result, it may be more acceptable to use `collect()` instead
+            // of `first()` to return all the possible requested fields.
+            .collect::<Vec<serde_json::Value>>()
+            .pop();
```
cobward: If it's possible that more than one paths match then we should return all of those values. The library needs to be as flexible and correct as possible. We can decide what to do with the result in our implementations downstream.
```suggestion
            .collect::<Vec<serde_json::Value>>();
```

src/core/input_descriptor.rs
```diff
@@ -662,67 +660,57 @@ impl ConstraintsField {
...
+
+        let purpose = self.purpose().map(ToOwned::to_owned);
+
+        let name = self
+            .name()
+            .map(ToOwned::to_owned)
+            // TODO: Add an "unknown field" if the name is not provided.
+            // Consider skipping or erroring on unknown fields.
+            .unwrap_or_default();
```
cobward: No need to lose information. If the request doesn't include a name then propagate that Option into the RequestedField struct
```suggestion
        let name = self
            .name()
            .map(ToOwned::to_owned);
```

cobward (changes requested):

cobward (approved):

PR #35, opened 2024-10-24
Improve submission validation and input extraction.

timothee-haudebourg:
This PR improves the presentation submission validation in two ways:

- Input extraction: currently we are only doing validation. But I also needed to extract the inputs provided in the submission. I added a `PresentationSubmission::find_inputs` function to find and extract the inputs. It returns a `MatchingInputs` structure containing all the inputs that matched the presentation definition.
- Input decoding: currently the presentation submission inputs are not decoded, which is required to correctly handle nested descriptor maps. I've added a `ClaimsDecoder` trait that is given to the `find_inputs` to decode the visited input claims. I provided a "default" implementation `NoClaimsDecoder` that doesn't decode anything (so it behaves exactly like the current implementation).

For the validation part:
- I added a `MatchingInput::validate` to validate the extracted input against the presentation definition requirements.
- I replaced the previous validation function with `PresentationSubmission::find_and_validate_inputs` that just calls `find_inputs` and `validate`.

## Other changes

I've set some fields public and removed some accessors.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,8 @@ http = "1.1.0"
 # This is currently used in the jwt_vp test to go from a `VeriableCredential` to an `AnyJsonCredential` type.
 # There may be a better way to handle this that doesn't require the `json-syntax` crate directly.
 json-syntax = { version = "0.12.5", features = ["serde_json"] }
-jsonpath_lib = "0.3.0"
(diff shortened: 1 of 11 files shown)


Ryanmtate (approved):
Nice work @timothee-haudebourg!

PR #39, opened 2024-10-31
return application json encoded authorization response

Ryanmtate:


src/core/response/mod.rs
```diff
@@ -26,45 +22,46 @@ impl AuthorizationResponse {
...
+            .context("failed to encode presentation_submission as JSON")?;
+
+        map.insert("presentation_submission", presentation_submission);
+        map.insert("vp_token", vp_token);
+
+        let encoded = serde_urlencoded::to_string(&map).context(
+            "failed to encode presentation_submission as 'application/x-www-form-urlencoded'",
+        )?;
+
+        Ok(encoded)
```
timothee-haudebourg, line 62: It works so I'm not going to block it but:
- there is no reason for the serialization to fail, so you can safely `unwrap` instead of returning a `Result`
- you could use a custom `struct` instead of a `HashMap` to avoid memory allocations.

timothee-haudebourg (approved):

src/core/response/mod.rs
Ryanmtate: commit [ad3974c](https://github.com/spruceid/openid4vp/pull/39/commits/ad3974c5d19c94ecefb4b83427886ff1c480c194) removes the hash map and uses a custom struct here instead.

Juliano1612 (approved):

PR #40, opened 2024-10-31
Feat/support unencoded jwt authorization requests

Joey-Silberman:
This PR adds support for unencoded/unsigned jwt authorization requests to bypass KID and signature checks in request validation for when `client_id_scheme` is `did`. This is necessary for using VC playground verifier QRs because their `request_jwt` is unsigned and thus doesn't have a KID.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ http = "1.1.0"
 # There may be a better way to handle this that doesn't require the `json-syntax` crate directly.
 json-syntax = { version = "0.12.5", features = ["serde_json"] }
 # jsonpath_lib = "0.3.0"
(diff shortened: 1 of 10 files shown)


Ryanmtate:
Changes in SSI have resolved this.

Ryanmtate:
This short circuit for avoiding the check whether the client id of the verifier (DID) in relation to the key ID in the header may still be helpful.

However, this may not be the correct way to handle whether or not a `kid` exists. This check makes the assumption that if the `alg` value is missing in the header, so will the `kid`. Is this valid? Do we need to handle this differently?

Ryanmtate (approved):

PR #42, opened 2024-11-13
add data integrity conversion for vp token

Ryanmtate:
Additionally adds a check for authorization request vp formats supported to check cryptosuite against
expected response formats.

### DIFF
--- a/src/core/authorization_request/verification/did.rs
+++ b/src/core/authorization_request/verification/did.rs
@@ -43,7 +43,6 @@ pub async fn verify_with_resolver(
     // This bypass is for unencoded JWT requests, but we will need to change this later
     // so that trust is preserved when receiving unencoded requests
     if alg.contains("none") {
-        // bail!("UNSIGNED JWT")
         return Ok(());
     }
 
--- a/src/core/authorization_request/verification/did.rs
+++ b/src/core/authorization_request/verification/did.rs
@@ -40,6 +40,13 @@ pub async fn verify_with_resolver(
         bail!("request was signed with unsupported algorithm: {alg}")
     }
 
+    // This bypass is for unencoded JWT requests, but we will need to change this later
+    // so that trust is preserved when receiving unencoded requests
+    if alg.contains("none") {
+        // bail!("UNSIGNED JWT")
+        return Ok(());
+    }
+
     let Json::String(kid) = headers
         .remove("kid")
         .context("'kid' was missing from jwt headers")?
--- a/src/core/authorization_request/verification/did.rs
+++ b/src/core/authorization_request/verification/did.rs
@@ -40,6 +40,14 @@ pub async fn verify_with_resolver(
         bail!("request was signed with unsupported algorithm: {alg}")
     }
 
+    // This bypass is for unencoded JWT requests, but we will need to change this later
+    // so that trust is preserved when receiving unencoded requests
+    // NOTE: This requires that `Algorithm::None` is permitted in the wallet metadata
+    // Otherwise, this function will error in the previous assertion.
+    if alg.contains("none") {
+        return Ok(());
+    }
+
     let Json::String(kid) = headers
         .remove("kid")
         .context("'kid' was missing from jwt headers")?
--- a/src/core/authorization_request/verification/did.rs
+++ b/src/core/authorization_request/verification/did.rs
@@ -40,12 +40,6 @@ pub async fn verify_with_resolver(
         bail!("request was signed with unsupported algorithm: {alg}")
     }
 
-    // This bypass is for unencoded JWT requests, but we will need to change this later
-    // so that trust is preserved when receiving unencoded requests
-    if alg.contains("none") {
-        return Ok(());
-    }
-
     let Json::String(kid) = headers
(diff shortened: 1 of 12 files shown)


src/core/authorization_request/verification/did.rs
```diff
@@ -40,6 +40,14 @@ pub async fn verify_with_resolver(
...
     }
 
+    // This bypass is for unencoded JWT requests, but we will need to change this later
+    // so that trust is preserved when receiving unencoded requests
+    // NOTE: This requires that `Algorithm::None` is permitted in the wallet metadata
+    // Otherwise, this function will error in the previous assertion.
+    if alg.contains("none") {
+        return Ok(());
+    }
+
```
Ryanmtate, line 50: According to the spec, I think we should check whether the `kid` exists before we check for `alg` is set to `none`.

See: https://datatracker.ietf.org/doc/rfc9101/ 

> The authorization server MUST validate the signature of the JWS-
   signed [RFC7515] Request Object.  If a "kid" Header Parameter is
   present, the key identified MUST be the key used and MUST be a key
   associated with the client.  The signature MUST be validated using a
   key associated with the client and the algorithm specified in the
   "alg" Header Parameter.  Algorithm verification MUST be performed, as
   specified in Sections 3.1 and 3.2 of [RFC8725].

>   If the key is not associated with the client or if signature
   validation fails, the authorization server MUST return an
   "invalid_request_object" error to the client in response to the
   authorization request.

cobward: I think [this requirement](https://openid.net/specs/openid-4-verifiable-presentations-1_0-20.html#section-5.7-6.2.1) from openid4vp supercedes this requirement:

> Since DID Document may include multiple public keys, a particular public key used to sign the request in question **MUST be identified by the kid in the JOSE Header**.

src/core/metadata/parameters/verifier.rs
```diff
@@ -8,6 +9,31 @@ use serde_json::{Map, Value as Json};
...
+    ///
+    /// NOTE: This method is interested in the cryptosuite of the claim format
+    /// payload and not the claim format designation.
+    ///
+    /// For example, the cryptosuite would need to match one of the `alg`
+    /// values in the claim format payload.
+    pub fn supports_cryptosuite(
+        &self,
+        format: &ClaimFormatDesignation,
+        cryptosuite: &String,
```
cobward: There's some mixed semantics here. Different formats use either `proof types` (AKA `cryptosuite`) or `alg`  (AKA signature algorithm) to identify how the credential/presentation is secured. Conflating them all as `cryptosuite` is confusing as that is specifically a W3C VC term.
```suggestion
    /// Returns a boolean to denote whether a particular pair of format and security method
    /// are supported in the VP formats. A security method could be a JOSE algorithm, a COSE
    /// algorithm, a Cryptosuite, etc.
    ///
    /// NOTE: This method is interested in the security method of the claim format
    /// payload and not the claim format designation.
    ///
    /// For example, the security method would need to match one of the `alg`
    /// values in the claim format payload.
    pub fn supports_security_method(
        &self,
        format: &ClaimFormatDesignation,
        security_method: &String,
```

cobward (changes requested):

src/core/authorization_request/verification/did.rs
Ryanmtate: I agree this requirement takes precedence. The question I have now, though, is that it seems the missing `kid` value in the header is in nonconformance with the oid4vp specification.

We are currently bypassing the `kid` check IFF the wallet metadata supports a `None` algorithm for signing.

This preserves the explicit check `kid`, but allows the wallet metadata to control bypassing this assertion.

@cobward What suggested changes do you think we should make here?

@Joey-Silberman anything you want to add?

cobward: I would suggest don't make the changes here. This library implements DID client id scheme request verification in the standard way. If we need to have a workaround for a particular integration then we can write a non-standard implementation of the RequestVerifier trait: https://github.com/spruceid/openid4vp/blob/main/src/core/authorization_request/verification/mod.rs#L29

src/core/metadata/parameters/verifier.rs
```diff
@@ -8,6 +9,34 @@ use serde_json::{Map, Value as Json};
...
+    ///
+    /// NOTE: This method is interested in the security method of the claim format
+    /// payload and not the claim format designation.
+    ///
+    /// For example, the security method would need to match one of the `alg`
+    /// values in the claim format payload.
+    pub fn supports_security_method(
+        &self,
+        format: &ClaimFormatDesignation,
+        security_method: &String,
```
cobward, line 25:
```suggestion
        security_method: &str,
```

src/core/authorization_request/verification/did.rs
cobward: If time doesn't permit making the update elsewhere (in mobile-sdk), then I don't think we should block merging this in. Just ensure that we clarify the issue with the relevant parties and fix this at the earliest opportunity.

cobward (approved):

PR #43, opened 2024-11-21
Feat/metadata improvements

Ryanmtate:
Adds Client Metadata accessor methods according to:

- https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-5.1-4.2.1
- https://openid.net/specs/oauth-v2-jarm-final.html#name-client-metadata


src/core/authorization_request/parameters.rs
```diff
@@ -169,6 +188,150 @@ impl ClientMetadata {
...
+
+    /// Return the `VpFormats` from the `client_metadata` field.
+    ///
+    /// vp_formats: REQUIRED when not available to the Wallet via another mechanism.
+    ///
+    /// As defined in [Section 10.1](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#client_metadata_parameters).
+    ///
+    /// See reference: https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-5.1-4.2.2.2
+    pub fn vp_formats(&self) -> Result<VpFormats, Error> {
+        self.0.get().ok_or(anyhow!("missing vp_formats"))?
```
cobward, line 224:
```suggestion
        self.0.get().parsing_error()?
```

src/core/authorization_request/parameters.rs
```diff
@@ -169,6 +188,150 @@ impl ClientMetadata {
...
+    ///
+    /// The jwks_uri or jwks metadata parameters can be used by clients to register their public encryption keys.
+    ///
+    /// See: https://openid.net/specs/oauth-v2-jarm-final.html#section-3-4
+    ///
+    pub fn jwks(&self) -> Option<Vec<JWK>> {
+        self.0 .0.get("jwks").and_then(|j| j.as_array()).map(|j| {
+            j.iter()
+                // NOTE: we're ignoring the errors here, but may wish to handle the errors in the future.
+                .filter_map(|jwk| serde_json::from_value(jwk.clone()).ok())
```
cobward: This and the other parameters that you have manually implemented are already implemented here:
https://github.com/spruceid/openid4vp/blob/main/src/core/metadata/parameters/verifier.rs

You should change these methods to `self.0.get()` as for vp_formats, or just do `client_metadata.0.get()` in the downstream code.

cobward (changes requested):

src/core/authorization_request/parameters.rs
Ryanmtate: Thanks for pointing this out. I updated to use the typed parameter implementations, and added missing typed parameter implementations for the other fields.

cobward (approved):

src/core/metadata/parameters/wallet.rs
```diff
@@ -205,6 +205,73 @@ impl From<AuthorizationEncryptionEncValuesSupported> for Json {
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct AuthorizationSignedResponseAlg(pub ssi::crypto::Algorithm);
+
```
cobward: These new values should be in `verifier.rs` not `wallet.rs`, right?

cobward (changes requested):
One final comment, otherwise LGTM

src/core/metadata/parameters/wallet.rs
Ryanmtate: Nice call, and I realize you already had some of these implemented.

I've removed the extraneous implementations in [204fc40](https://github.com/spruceid/openid4vp/pull/43/commits/204fc40606f0b2389fdf728d885d1649f406b58b)

cobward (approved):

PR #45, opened 2024-12-16
patch credential match validation for array of values

Ryanmtate:
Updates `is_credential_match` to check each value in an array of values.

### DIFF
--- a/src/core/presentation_definition.rs
+++ b/src/core/presentation_definition.rs
@@ -223,23 +223,16 @@ impl PresentationDefinition {
             .filter(|field| field.is_required())
             .all(|field| {
                 match field.filter.as_ref().map(|f| f.validator()) {
-                    Some(validator) => {
-                        let is_valid = field
-                            .path
-                            .iter()
-                            // NOTE: Errors are ignored to allow other paths to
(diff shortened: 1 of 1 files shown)


src/core/presentation_definition.rs
```diff
@@ -223,23 +223,16 @@ impl PresentationDefinition {
...
+                    Some(validator) => field
+                        .path
+                        .iter()
+                        .flat_map(|path| path.query(credential))
+                        .any(|value| match value {
+                            serde_json::Value::Array(vals) => {
+                                vals.iter().any(|val| validator.validate(val).is_ok())
+                            }
+                            _ => validator.validate(value).is_ok(),
+                        }),
```
cobward: Just realized this introduces a bug if the constraint _is_ for an array, so probably need to do this:

```suggestion
                        .any(|value| {
                            match value {
                                // If the value is an array, test the constraint against each array element.
                                serde_json::Value::Array(vals) => {
                                    if vals.iter().any(|val| validator.validate(val).is_ok()) {
                                        return true;
                                    }
                                },
                                _ => (),
                            }
                            validator.validate(value).is_ok()
                        }),
```

cobward (approved):

PR #53, opened 2025-01-17
Adds path to requested field

w4ll3:
Adds a new field to Requested Field in order to support selective disclosure

### DIFF
--- a/src/core/input_descriptor.rs
+++ b/src/core/input_descriptor.rs
@@ -126,7 +126,9 @@ impl InputDescriptor {
         self.constraints
             .fields
             .iter()
-            .map(|field| field.requested_fields(self.id.clone(), value))
+            .map(|field| {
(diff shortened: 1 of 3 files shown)


src/core/input_descriptor.rs
```diff
@@ -540,6 +540,7 @@ impl ConstraintsField {
         RequestedField {
             id: uuid::Uuid::new_v4(),
             name: self.name.clone(),
+            path: self.path.iter().map(|j| j.to_string()).collect(),
```
Ryanmtate: Is it possible to select only the path that matched the requested field?

IIRC, `self.path` in this instance is the _possible_ paths where the field may reside. One of these paths should match the constraint and we should use that path.

WDYT?

w4ll3: do we know that at `OID4VP`?

Ryanmtate (approved):

PR #54, opened 2025-01-24
[FIX] Remove extraneous quotations around url encoded vp_token

Ryanmtate:
This PR adds a helper method `clean_quoted_string` to strip extra quotations in JSON encoded `vp_token` payload for url encoding.

There might be a better way to serialize the `VpToken` to ensure an extra quotation is not included, open to suggestions.

Drive by changes:

- Adds an optional `state` property to from the authorization request to the authorization response.
- Adds some serde derive macros to help debug.

### DIFF
--- a/src/core/authorization_request/mod.rs
+++ b/src/core/authorization_request/mod.rs
@@ -1,7 +1,7 @@
 use std::ops::{Deref, DerefMut};
 
 use anyhow::{anyhow, bail, Context, Error, Result};
-use parameters::ClientMetadata;
+use parameters::{ClientMetadata, State};
 use serde::{Deserialize, Serialize};
 use serde_json::Value as Json;
 use url::Url;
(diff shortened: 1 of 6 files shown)


cobward (changes requested):
I'm not certain we should implement it this way. Older versions of the spec require the vp_token to be a JSON encoded string:
https://github.com/spruceid/openid4vp/blob/2210b2c1368f19f639f810f95b801e253c6e635d/src/core/response/parameters.rs#L38

The latest version of the spec removes the "JSON" from that requirement, which is why there is this mismatch between the implementation here and some implementations we see in the wild.

I'm worried that if we make this change, we will unknowingly break current and future implementations. Could you instead use a flag to decide whether or not to use JSON string encoding?

Ryanmtate:
> I'm worried that if we make this change, we will unknowingly break current and future implementations. Could you instead use a flag to decide whether or not to use JSON string encoding?

How would we determine this in practice? We know ahead of time with CA, but for other verifiers, hard to know whether it will be accepted or not.

cobward:
> > I'm worried that if we make this change, we will unknowingly break current and future implementations. Could you instead use a flag to decide whether or not to use JSON string encoding?
> 
> How would we determine this in practice? We know ahead of time with CA, but for other verifiers, hard to know whether it will be accepted or not.

We can only know by testing. The default behaviour can be to not include the quotation marks (as that is the future of the spec), but I think we need a way to include them. The quotation marks are in the current version because that is what was required for a previous integration.

cobward (approved):

PR #59, opened 2025-04-30
Implement support for DCQL and draft 24

cobward:


sbihel (approved):
I should have renamed the original `build` function to avoid confusion with `build_dc_api` -- but we'll see how the specs evolve

PR #65, opened 2025-08-28
fix: Resolve issue while retrieving auth endpoint url (#55)

### LINKED ISSUE #55: WalletMetadata:new() leads to error with AuthorizationEndpoint
I constructed WalletMetadata via the `new()` function. This function requires as first parameter the `AuthorizationEndpoint`, however when passing it to `RequestBuilder.build()` the method complains that the `AuthorizationEndpoint` is missing. As a workaround, I had to set the authorization endpoint in addition in the `WalletMetadata::new(other)` parameter as it is done in the examples. It seems like some data conversion is failing so that `wallet_metadata.get()` can't get a hold of the AuthorizationEndpoint parameter.

Abdulbois:
- Update approach of retrieving auth endpoint url while building authorization request
- Resolves #55

### DIFF
--- a/src/verifier/request_builder.rs
+++ b/src/verifier/request_builder.rs
@@ -164,9 +164,9 @@ impl<'a> RequestBuilder<'a> {
         };
 
         let authorization_endpoint = wallet_metadata
-            .get::<AuthorizationEndpoint>()
-            .parsing_error()?
-            .0;
+            .authorization_endpoint()
+            .0
+            .clone();
 
         let authorization_request_url = AuthorizationRequest {
             client_id: Some(client_id.0.clone()),
--- a/src/verifier/request_builder.rs
+++ b/src/verifier/request_builder.rs
@@ -10,11 +10,8 @@ use crate::{
             AuthorizationRequest, AuthorizationRequestObject, RequestIndirection,
         },
         dcql_query::DcqlQuery,
-        metadata::{
-            parameters::wallet::{AuthorizationEndpoint, ClientIdSchemesSupported},
-            WalletMetadata,
-        },
-        object::{ParsingErrorContext, TypedParameter, UntypedObject},
+        metadata::{parameters::wallet::ClientIdSchemesSupported, WalletMetadata},
+        object::{TypedParameter, UntypedObject},
         presentation_definition::PresentationDefinition,
     },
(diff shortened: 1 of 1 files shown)


Abdulbois:
@sbihel, please review.
As described in #55, [these](https://github.com/spruceid/openid4vp/blob/2b0c8bae7e0f00813f886e63adaccf635b931d0e/src/verifier/request_builder.rs#L166-L170)  lines cause an error when the `other`  field is None while building the `WalletMetadata` because the `UntypedObject` map will not contain the `authorization_endpoint` field.

cobward:
Can you provide a minimal example showing the issue this is resolving?

Abdulbois:
> Can you provide a minimal example showing the issue this is resolving?

Just for testing purposes, in order to bypass `build_with_session_id` validation, try to parse `authorization_endpoint` while calling `build` method of `RequestBuilder`

```
    pub async fn build(self, wallet_metadata: WalletMetadata) -> Result<(Uuid, Url)> {
        let uuid = Uuid::new_v4();
        let authorization_endpoint = wallet_metadata // Test parsing
            .get::<AuthorizationEndpoint>()
            .parsing_error()?;

        let authorization_request_url = self.build_with_session_id(uuid, wallet_metadata).await?;

        Ok((uuid, authorization_request_url))
    }
```

cobward:
I see, this change is fine, but it doesn't actually address the problem in #55. I think WalletMetadata needs a rethink.

cobward:
Actually I'm concerned that this could be a break some implementations that edit the inner UntypedObject, as now the request will be built with the original AuthorizationEndpoint even if the inner value was updated.

Abdulbois:
> I see, this change is fine, but it doesn't actually address the problem in #55. I think WalletMetadata needs a rethink.

Seems, current changes will at least fix that error and remove  that workaround described in #55

Totally agree with the point that `other` field of `WalletMetadata` should be handled correctly

Abdulbois:
> Actually I'm concerned that this could be a break some implementations that edit the inner UntypedObject, as now the request will be built with the original AuthorizationEndpoint even if the inner value was updated.

If I understood correctly, currently inner methods of `WalletMetadata` doesn't touch/update value of `authorization_endpoint` while using `inner`(`UntypedObject`) field, is not it?
Or you mean something else?

PR #76, opened 2026-05-01
fix: accept encryption JWKs without explicit `use` claim

Ryanmtate:
## Summary

Per OID4VP v1.0 §8.3, only `alg` is required on encryption JWKs in `client_metadata.jwks`; `use` is optional. The validation pass in `validate_request_against_metadata` (`src/core/authorization_request/verification/mod.rs:206-211`) was rejecting JWKs without an explicit `use:"enc"`, even though the runtime selector in `src/core/jwe.rs:202-206` already accepts them (it logs `JWK missing 'use' parameter, assuming it can be used for encryption` and proceeds).

This left the validation pass stricter than the actual encryption code, so requests the wallet would have successfully encrypted against were rejected upfront.

The fix flips `unwrap_or(false)` → `unwrap_or(true)` in the filter: a missing `use` is now treated as acceptable, while an explicit non-`enc` use (e.g. `sig`) still excludes the key. All 86 existing lib tests pass.

## How this was surfaced

Found while running the OIDF conformance suite's wallet test plans (`oid4vp-1final-wallet-haip-test-plan` and `oid4vp-1final-wallet-test-plan`) against a SpruceID-built wallet. The suite step `AddVP1FinalEncryptionParametersToClientMetadataWithoutUseEnc` **intentionally omits `use:"enc"`** in the encryption JWK to verify spec compliance.

With the unpatched lib, every test using `response_mode=dc_api.jwt` (or `direct_post.jwt`) failed identically at validation. The wallet-side error (captured via `adb logcat`):

```
com.spruceid.mobile.sdk.rs.DcApiException$InvalidRequest:
  v1=the request is could not be verified:
  unable to validate Authorization Request:
  no encryption key with 'alg' found in jwks (required per OID4VP v1.0 Section 8.3)
```

> **Note on suite-side observability**: the OIDF suite only sees the wallet's failure as a generic `Browser API threw an exception` / `NetworkError: Error retrieving a token` — the actual underlying error (`no encryption key with 'alg' found...`) is only visible in device-side wallet logs. So the linked conformance trace below shows the same outer symptom that any wallet rejection produces; the proof that this specific lib check is the culprit is in the verbatim Rust error message above, captured from the wallet's logcat at the moment of failure.

This check fires *before* any `client_id_prefix` dispatch in `verify_request`, so it shadowed both `x509_san_dns` and `x509_hash` happy-flow runs — neither code path was even reached. Once this lands, those signed-request paths can be exercised end-to-end against the conformance suite.

Published failure trace (read-only, no auth) for an `iso_mdl + dc_api.jwt + x509_hash + request_uri_signed + haip` happy-flow run on `demo.certification.openid.net`:

- Test: https://demo.certification.openid.net/log-detail.html?log=uvSNB4soyDXfEHy&public=true
- Plan: https://demo.certification.openid.net/plan-detail.html?plan=kAkWzRKQM1U8A&public=true

## Test plan

- [x] `cargo build` clean
- [x] `cargo test --lib` — 86 passed, 0 failed
- [ ] Re-run the OIDF wallet plan (`iso_mdl + dc_api.jwt + x509_hash`) against a wallet rebuilt with this fix and confirm the run progresses past `validate_request_against_metadata` (currently blocked on a separate wallet-side rebuild step)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,7 +63,7 @@ sha2 = "0.10"
 thiserror = "1.0"
 tokio = { version = "1.32.0", features = ["macros", "full"] }
 tower = "0.5"
-tower-http = { version = "0.6", features = ["cors", "trace", "timeout", "limit"] }
+tower-http = { version = "0.6.7", features = ["cors", "trace", "timeout", "limit"] }
 tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter", "json"] }
 url = { version = "2.5", features = ["serde"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ serde_json = "1.0.107"
 serde_json_path = "0.7.1"
 serde_urlencoded = "0.7.1"
 sha2 = "0.10"
-ssi = { version = "0.15", features = ["secp256r1"] }
+ssi = { version = "0.16", features = ["secp256r1"] }
 thiserror = "1.0.65"
 tokio = "1.32.0"
 tracing = "0.1.37"
--- a/examples/verifier-conformance-adapter/server/mod.rs
+++ b/examples/verifier-conformance-adapter/server/mod.rs
@@ -27,8 +27,5 @@ pub fn create_router(state: Arc<AppState>) -> Router {
         .with_state(state)
         .layer(CorsLayer::permissive())
         .layer(TraceLayer::new_for_http())
-        .layer(TimeoutLayer::with_status_code(
-            axum::http::StatusCode::REQUEST_TIMEOUT,
-            Duration::from_secs(30),
-        ))
+        .layer(TimeoutLayer::new(Duration::from_secs(30)))
 }
--- a/examples/verifier-conformance-adapter/server/mod.rs
+++ b/examples/verifier-conformance-adapter/server/mod.rs
@@ -27,5 +27,8 @@ pub fn create_router(state: Arc<AppState>) -> Router {
         .with_state(state)
         .layer(CorsLayer::permissive())
         .layer(TraceLayer::new_for_http())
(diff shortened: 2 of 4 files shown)


Ryanmtate:
core2 v0.4 has been yanked: https://crates.io/crates/core2 and the repo archived. This is not a direct dependency:


```
cargo tree -i core2                                                                                                                                                                                  

core2 v0.4.0
├── cid v0.8.6
│   └── libipld-core v0.14.0
│       ├── libipld v0.14.0
│       │   └── ssi-ucan v0.4.0
│       │       └── ssi v0.15.0
│       │           └── openid4vp v0.1.0 (/Users/rmt/Projects/SpruceId/openid4vp)
│       ├── libipld-cbor v0.14.0
│       │   └── libipld v0.14.0 (*)
│       ├── libipld-json v0.14.0
│       │   └── libipld v0.14.0 (*)
│       └── libipld-macro v0.14.0
│           └── libipld v0.14.0 (*)
├── libipld-core v0.14.0 (*)
└── multihash v0.16.3
    ├── cid v0.8.6 (*)
    ├── libipld v0.14.0 (*)
    ├── libipld-core v0.14.0 (*)
    └── libipld-json v0.14.0 (*)
```

Ryanmtate:
https://github.com/multiformats/rust-cid/commit/54855bab2cbcecd20ef6b1c59c73d273c9077371

sbihel (commented):
You can't let any JWK without a `use` field be used for encryption because some algorithms (e.g. `ES256`) are only for signing. You need to do another filtering based on what algorithms can be used for encryption.

I know that there's another round of filtering based on the wallet's supports algs, but it's not guaranteed to run.

Could you also add some tests please?

PR #77, opened 2026-05-30
Update conformance tests according to OIDF OID4VP 1.0/HAIP 1.0 tests

Juliano1612:
This updates the code as follows:

- Verifier and wallet conformance adapters gain full HAIP support: `x509_hash` client ID, CA-signed (non-self-signed) leaf with `x5c`, signed request verification, encrypted `direct_post.jwt` responses, real SD-JWT VC issuance, claim/data-minimization handling, and the negative tests (invalid signature, unknown prefix, unknown transaction_data type).

- Two spec-compliance fixes in the library, surfaced and fixed while running the OpenID conformance suite:
  - §8.3: request validation no longer requires `use: enc` on the verifier's encryption JWK. The spec only mandates `alg`; `use` is an optional selection hint. Now aligned with `find_encryption_jwk`.
  - §5.8: the request builder now defaults the request object `aud` to `https://self-issued.me/v2` (static discovery), overridable for dynamic discovery.

### DIFF
--- a/examples/verifier-conformance-adapter/server/handlers.rs
+++ b/examples/verifier-conformance-adapter/server/handlers.rs
@@ -17,6 +17,7 @@ use openid4vp::{
 use serde::{Deserialize, Serialize};
 use ssi::claims::jws::{decode_unverified, decode_verify};
 use ssi::claims::sd_jwt::{KbJwtPayload, SdAlg, SdJwt};
+use ssi::claims::{DateTimeProvider, ValidateClaims};
 use ssi::jwk::JWK;
 use tracing::{debug, error, info};
 use uuid::Uuid;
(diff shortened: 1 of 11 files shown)


src/core/authorization_request/verification/mod.rs
```diff
@@ -199,16 +199,11 @@ pub(crate) async fn validate_request_against_metadata<W: Wallet + ?Sized>(
...
         let encryption_algs: Vec<String> = jwks
             .keys
             .iter()
-            .filter(|k| {
-                k.get("use")
-                    .and_then(|v| v.as_str())
-                    .map(|s| s == "enc")
-                    .unwrap_or(false)
-            })
+            .filter(|k| !matches!(k.get("use").and_then(|v| v.as_str()), Some(u) if u != "enc"))
```
sbihel: I'm having a hard time understanding this filtering -- isn't doing the same as it did before?

I brought this up in #76, but I don't think accepting any JWK without a specified `use` would work. `find_encryption_jwk` at least does a check on the `crv` to ensure it would be compatible for encryption. So I think maybe this block should use the `find_encryption_jwk` function, even though it's not perfect and should cover more `crv`/`kty`/`alg` (which can be done at a later time)

examples/verifier-conformance-adapter/server/handlers.rs
```diff
@@ -219,11 +200,183 @@ pub async fn receive_response(
...
+        return Err("KB-JWT sd_hash does not match the presented SD-JWT".into());
+    }
+
+    // The KB-JWT must be fresh: `iat` cannot be in the future (beyond clock
+    // skew) nor too far in the past. The window is intentionally generous; tune
+    // it if the conformance suite uses a smaller offset.
+    const CLOCK_SKEW_SECS: f64 = 60.0;
+    const MAX_AGE_SECS: f64 = 300.0;
+    let iat = kb.iat.0.as_seconds();
+    let now = std::time::SystemTime::now()
```
sbihel, line 338: isn't this something that `ssi` should be doing already? Just to make sure we're testing as much of our libraries as possible

sbihel (approved):
I haven't reviewed the conformance adapters in detail as they're inherently tested by the conformance suite

src/core/authorization_request/verification/mod.rs
Juliano1612: Switched to `find_encryption_jwk` as suggested. Two notes on the "same as before?" question:

It's not the same. The old filter (`use == "enc"`, else drop) excluded keys with no `use`. The new predicate includes them and only drops keys whose `use` is explicitly non-`enc`. That was the point of the §8.3 fix: the spec selects the key from `kty`/`crv`/`alg` and only `alg` MUST be present. `use` is just an optional hint. The conformance test sends an `enc` JWK with no `use`, which the old filter dropped.

But you're right about the gap. With `use` optional, blindly reading `alg` could pick up a signing key's alg. `find_encryption_jwk` guards that via its `crv`/`alg` checks and is the same selection `build_encrypted_response` uses, so validation now matches actual response building instead of duplicating a looser version. 

commit https://github.com/spruceid/openid4vp/pull/77/changes/d1fc400e76bef6200247af18f1017bf4bc670baa

examples/verifier-conformance-adapter/server/handlers.rs
Juliano1612: Good catch! `ssi` does validate `iat`. Two things `ssi` doesn't cover, and we still need to implement in the adapter side: 
1) `ssi`'s `iat` check is strict (`iat <= now`, no clock-skew leeway)
2) It has no "too far in the past" / max-age check. That's verifier policy, not something the library decides. So the `iat in the future` test goes through `ssi`, and the `iat in the past` test is enforced here.

commit https://github.com/spruceid/openid4vp/pull/77/commits/eb19c3991330ff21e27093252bd8ec078c56da65

